Issue #1, opened 2022-12-01
As Vaults user, I need an informative landing page when I first navigate to vaults UI


otoole-brendan:
## What is the Problem Being Solved?

We need to implement a new Vaults landing page that serves as a informational jumping off point for users to proceed in creating a vault. It's intended to explain to users at a glance what vaults are/what they can do and funnel them to clear 'Create Vault' CTA.

## Description of the Design

-Reference Figma design https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=1876%3A14385&t=M1K9MKcyPF9GJnCO-0 (although note Ekko need to convey the tiles arent clickable)

- [ ] footer links including Inter Discord, docs, Twitter etc.
- [ ] clickable analytics button which brings users to analytics.inter.trade

## Security Considerations
Unknown

## Test Plan
Smoke Test upon implementation

otoole-brendan:
@turadg @samsiegart  The purpose of this page has evolved - it is not just to show vault type previews, it serves as informative vault landing page where we they can learn at a glance what they can do/what vaults are and click a CTA button to start creating a vault. In short it's more of a marketing/informative jumping off point so should not be blocked by us not supporting multiple vault types at launch.
I'd like to remove the dependency so that is clear, I will also rename this issue so it's clearer.

otoole-brendan:
Edited ticket title per my comment

otoole-brendan:
@samsiegart I have added more specific detail/requirements based on latest Ekko design - some are not MVP critical, I have marked as such

otoole-brendan:
Note: in an effort to cut scope, this view has been deferred. Instead, users will land on My Vaults Dashboard page. General page requirements have been moved to this ticket https://github.com/Agoric/dapp-inter/issues/12

otoole-brendan:
Not needed

PR #24, opened 2022-12-16
View collateral choices

samsiegart:
refs https://github.com/Agoric/dapp-inter/issues/1

For now it just queries vstorage for all the vault manager keys, e.g. `published.vaultFactory.manager0`, from there we can read their metrics and governed params etc.

Uses `react-query` to query vstorage, since `casting` only supports querying data not keys.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -21,6 +21,6 @@ export const networkConfigs = {
   },
   localhost: {
     label: 'Local Network',
-    url: 'http://localhost:3000/wallet/network-config',
+    url: 'http://wallet.agoric.app/wallet/network-config',
   },
 };
--- a/src/store/app.ts
+++ b/src/store/app.ts
@@ -1,0 +1,7 @@
+import { networkConfigs } from 'config';
+import { atomWithStorage } from 'jotai/utils';
+
+export const networkConfigAtom = atomWithStorage(
+  'agoric-network-config',
+  networkConfigs.mainnet,
+);
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,3 +1,4 @@
+// Until casting supports querying keys https://github.com/Agoric/agoric-sdk/issues/6690
 export const fetchVstorageKeys = (
   rpcAddr: string,
   path: string,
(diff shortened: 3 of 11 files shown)


src/store/app.ts
```diff
@@ -0,0 +1,7 @@
+import { networkConfigs } from 'config';
+import { atomWithStorage } from 'jotai/utils';
```
turadg, line 2: I thought we were moving away from Jotai. This WithStorage utility is pretty handy tho

src/util.ts
```diff
@@ -0,0 +1,33 @@
+export const fetchVstorageKeys = (
```
turadg, line 1: please note this issue to provide key queries in casting: https://github.com/Agoric/agoric-sdk/issues/6690

src/views/ErrorPage.tsx
```diff
@@ -0,0 +1,20 @@
+import { useRouteError } from 'react-router-dom';
+
+type RouteError = { statusText: string; message: string };
+
+const ErrorPage = () => {
+  const error = useRouteError() as RouteError;
+  console.error(error);
```
turadg, line 7: consider prefacing the output to indicate that it's coming from this component

src/util.ts
```diff
@@ -0,0 +1,33 @@
...
+    );
+};
+
+export const fetchRPCAddr = async (netconfigURL: string) => {
+  const response = await fetch(netconfigURL, {
+    headers: { accept: 'application/json' },
+  });
+  const networkConfig = await response.json();
+
+  return networkConfig.rpcAddrs[0];
```
turadg, line 32: why just the first? best to round-robin, no?

should casting abstract this too?

turadg (approved):

src/config.ts
```diff
@@ -21,6 +21,6 @@ export const networkConfigs = {
   },
   localhost: {
     label: 'Local Network',
-    url: 'http://localhost:3000/wallet/network-config',
+    url: 'http://wallet.agoric.app/wallet/network-config',
```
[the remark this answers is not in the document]
samsiegart: The wallet hosts a net config that points to the local chain, I didn't want to have to run a local wallet ui just to access the local chain.

src/store/app.ts
samsiegart: I'd prefer not to and don't recall that decision, I think it will be useful going forward but we can discuss more.

src/util.ts
samsiegart: Left a comment

src/views/ErrorPage.tsx
samsiegart: Done

src/util.ts
samsiegart: Just changed it to use a random selection. We could get fancier, but at least across many clients this will even the load. I think all the netconfigs only use a single rpc anyway though.

Casting does handle this, so if we could query the keys with casting we could remove this too.

PR #27, opened 2023-01-03
fix: update price feed data format

samsiegart:
Also added a button that will eventually be used to test sending an loan offer.

### DIFF
--- a/src/components/NewVault.tsx
+++ b/src/components/NewVault.tsx
@@ -1,0 +1,34 @@
+import { useVaultStore } from 'store/vaults';
+
+const NewVault = ({ id }: { id: string }) => {
+  const {
+    vaultGovernedParams,
+    vaultLoadingErrors,
+    vaultManagers,
+    vaultMetrics,
+    prices,
+    priceErrors,
+  } = useVaultStore();
+
+  const manager = vaultManagers.get(id);
+  const metrics = vaultMetrics.get(id);
+  const params = vaultGovernedParams.get(id);
+  const brand = metrics?.retainedCollateral?.brand;
+  const price = brand && prices.get(brand);
+
+  const error = vaultLoadingErrors.get(id) || (brand && priceErrors.get(brand));
+
+  const isLoading =
+    !(manager && metrics && params && price) || error != undefined;
+
+  return (
+    <button
+      className="bg-purple-500 hover:bg-purple-700 text-white font-semibold py-2 px-4 rounded"
+      disabled={isLoading}
+    >
+      Borrow 10 IST
+    </button>
+  );
+};
+
+export default NewVault;
--- a/src/components/NewVault.tsx
+++ b/src/components/NewVault.tsx
@@ -1,6 +1,8 @@
 import { useVaultStore } from 'store/vaults';
 
-const NewVault = ({ id }: { id: string }) => {
+type Props = { id: string };
+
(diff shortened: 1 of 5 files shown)


src/components/NewVault.tsx
```diff
@@ -0,0 +1,34 @@
+import { useVaultStore } from 'store/vaults';
+
+const NewVault = ({ id }: { id: string }) => {
```
turadg, line 3: please define `Props` type or interface. It's a good consistency for easy reading. 
```suggestion
const NewVault = ({ id }: Props) => {
```

src/components/NewVault.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const manager = vaultManagers.get(id);
+  const metrics = vaultMetrics.get(id);
+  const params = vaultGovernedParams.get(id);
+  const brand = metrics?.retainedCollateral?.brand;
+  const price = brand && prices.get(brand);
+
+  const error = vaultLoadingErrors.get(id) || (brand && priceErrors.get(brand));
+
+  const isLoading =
+    !(manager && metrics && params && price) || error != undefined;
```
turadg, line 22: consider simplifying the logical expression:
```suggestion
  const isReady =
    manager && metrics && params && price && error === undefined;
```

src/store/vaults.ts
```diff
@@ -6,7 +6,11 @@ export type Ratio = {
   denominator: Amount<'nat'>;
 };
 
-export type QuoteAmount = Amount<'set'>;
+export type PriceQuote = {
+  amountIn: Amount<'nat'>;
+  amountOut: Amount<'nat'>;
+  timestamp: bigint;
+};
```
turadg, line 13: [priceAggregatorChainlink publishes `PriceDescription`](https://github.com/Agoric/agoric-sdk/blob/c542cb6eaa52b131c589ad5197373276db6c2452/packages/inter-protocol/src/priceAggregatorChainlink.js#L142) so please import [that](https://github.com/Agoric/agoric-sdk/blob/c542cb6eaa52b131c589ad5197373276db6c2452/packages/zoe/tools/types.js#L16) instead

turadg (changes requested):

src/components/NewVault.tsx
samsiegart: Done

src/store/vaults.ts
samsiegart: I added `import type { PriceDescription } from '@agoric/zoe/tools/types';` and get `File '/home/samsiegart/dapp-inter/node_modules/@agoric/zoe/tools/types.d.ts' is not a module.`.

I went into types.d.ts and changed `type PriceDescription = {...}` to `export type PriceDescription = {...}` locally and it worked. How can we fix the sdk to export the types by default?

turadg: Shoot. Well 5f9b5df44822bb794ac219ba00e1f41aa6c56e16 adds a hack, good enough to merge.

> How can we fix the sdk to export the types by default?

That would be more of a mess, creating a lot of surface that can break.

Since the types are structural, it's not horrible to copy them over. But when we do it's use the same name as in agoric-sdk and make a comment pointing to the original. Those can be culled and cleaned up at some point. (I'm not tackling that right now because prices probably don't below in Zoe anyway.)

turadg (approved):

Issue #11, opened 2022-12-01
As Vaults user creating a new vault, once I have confirmed vault configuration I need to be able to ‘Make/Submit Offer’ or ‘Cancel’ in the ‘Offer Summary’ tile


otoole-brendan:
## What is the Problem Being Solved?
User needs ability to cancel or submit/make offer

## Description of the Design

"Cancel' button visible on 'Configure' box. Action should take undo selection/reset view.

Button placement as per https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=1938%3A13350&t=I65DB2Ve9pmj26hG-0

## Security Considerations
Unknown
## Test Plan
Smoke Test after implementation

PR #29, opened 2023-01-11
Open vault

samsiegart:
refs https://github.com/Agoric/dapp-inter/issues/11

### DIFF
--- a/src/components/CollateralChoice.tsx
+++ b/src/components/CollateralChoice.tsx
@@ -12,6 +12,8 @@ const CollateralChoice = ({ id }: { id: string }) => {
     vaultMetrics,
     prices,
     priceErrors,
+    vaultFactoryParams,
+    vaultFactoryParamsLoadingError,
   } = useVaultStore();
 
   const purses = useAtomValue(pursesAtom);
@@ -21,7 +23,10 @@ const CollateralChoice = ({ id }: { id: string }) => {
   const brand = metrics?.retainedCollateral?.brand;
   const price = brand && prices.get(brand);
 
-  const error = vaultLoadingErrors.get(id) || (brand && priceErrors.get(brand));
+  const error =
+    vaultLoadingErrors.get(id) ||
+    (brand && priceErrors.get(brand)) ||
+    vaultFactoryParamsLoadingError;
   const displayFunctions = useAtomValue(displayFunctionsAtom);
   if (error || !displayFunctions) {
     return (
@@ -35,7 +40,13 @@ const CollateralChoice = ({ id }: { id: string }) => {
     );
   }
 
-  const isLoading = !(manager && metrics && params && price);
+  const isLoading = !(
+    manager &&
+    metrics &&
+    params &&
+    price &&
+    vaultFactoryParams
+  );
   const {
     displayAmount,
     displayBrandPetname,
@@ -110,7 +121,11 @@ const CollateralChoice = ({ id }: { id: string }) => {
         {displayAmount(metrics.totalShortfallReceived)}{' '}
         {displayBrandPetname(metrics.totalShortfallReceived.brand)}
       </p>
-      <p>Purse Balance: {purseBalance}</p>
+      <p>Purse balance: {purseBalance}</p>
+      <p>
+        Minimum initial debt: {displayAmount(vaultFactoryParams.minInitialDebt)}{' '}
+        {displayBrandPetname(vaultFactoryParams.minInitialDebt.brand)}
+      </p>
       <NewVault id={id} />
     </>
   );
--- a/src/components/CollateralChoice.tsx
+++ b/src/components/CollateralChoice.tsx
@@ -40,13 +40,7 @@ const CollateralChoice = ({ id }: { id: string }) => {
     );
   }
 
-  const isLoading = !(
-    manager &&
-    metrics &&
-    params &&
-    price &&
-    vaultFactoryParams
-  );
+  const isReady = manager && metrics && params && price && vaultFactoryParams;
   const {
     displayAmount,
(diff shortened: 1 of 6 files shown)


src/components/NewVault.tsx
```diff
@@ -10,24 +13,53 @@ const NewVault = ({ id }: Props) => {
...
+  );
+
+  const IST_TO_BORROW = 5_000_000n; // 5 IST
 
-  const isReady = manager && metrics && params && price && error === undefined;
+  // TODO: Calculate these based on user inputs and actual prices and liquidation ratios.
+  const proposeOffer = () => {
+    makeOpenVaultOffer(
+      collateralPurse?.pursePetname,
+      10_000_000_000n,
```
turadg, line 50: curious why this value is inlined in `5_000_000n` got a constant

src/components/NewVault.tsx
```diff
@@ -10,24 +13,53 @@ const NewVault = ({ id }: Props) => {
...
+    p => p.brand === metrics?.retainedCollateral?.brand,
+  );
+
+  const IST_TO_BORROW = 5_000_000n; // 5 IST
 
-  const isReady = manager && metrics && params && price && error === undefined;
+  // TODO: Calculate these based on user inputs and actual prices and liquidation ratios.
+  const proposeOffer = () => {
+    makeOpenVaultOffer(
+      collateralPurse?.pursePetname,
```
turadg, line 49: `makeOpenVaultOffer` says the petname params are `string`. Why is it okay to pass undefined?

src/utils/displayFunctions.ts
```diff
@@ -64,11 +64,18 @@ export const makeDisplayFunctions = (brandToInfo: Map<Brand, BrandInfo>) => {
   const displayBrandIcon = (brand?: Brand | null) =>
     getLogoForBrandPetname(getPetname(brand));
 
-  const displayPrice = (price: PriceQuote) => {
-    return '$' + stringifyValue(price.amountOut.value, AssetKind.NAT, 2);
+  const displayPrice = (price: PriceDescription) => {
+    const PRICE_BRAND_UNIT_AMOUNT = 1_000_000n;
+    const USD_BRAND_DECIMALS = 6;
```
turadg, line 69: these constants should be module-scope and well documented as to why they're necessary.

samsiegart: done

src/components/NewVault.tsx
samsiegart: I wonder why I didn't see a type error there. This is technically a bug, worst case you see faulty proposal show up in your wallet, but it would require unlucky timing for a user to get there. Fixed with an assert, this is just temporary for demo anyway.

src/components/NewVault.tsx
```diff
@@ -10,24 +13,53 @@ const NewVault = ({ id }: Props) => {
...
+      IST_TO_BORROW,
+    );
+  };
 
   return (
     <button
-      className="bg-purple-500 hover:bg-purple-700 text-white font-semibold py-2 px-4 rounded"
+      className="disabled:bg-gray-300 disabled:cursor-not-allowed bg-purple-500 hover:bg-purple-700 text-white font-semibold py-2 px-4 rounded"
       disabled={!isReady}
+      onClick={() => proposeOffer()}
```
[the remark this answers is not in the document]
samsiegart: It does work but that way passes the event arg to `proposeOffer`, which I thought would be cleaner to explicitly leave out. In the past with class components you had to be more careful though https://reactjs.org/docs/faq-functions.html

src/components/NewVault.tsx
samsiegart: No rhyme or reason, fixed

src/components/NewVault.tsx
```diff
@@ -10,24 +13,53 @@ const NewVault = ({ id }: Props) => {
...
+    !error &&
+    offerSigner.isDappApproved;
+
+  const purses = useAtomValue(pursesAtom);
+  const istPurse = purses?.find(p => p.brand === metrics?.totalDebt?.brand);
+  const collateralPurse = purses?.find(
+    p => p.brand === metrics?.retainedCollateral?.brand,
+  );
+
+  const IST_TO_BORROW = 5_000_000n; // 5 IST
```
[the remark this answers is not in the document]
samsiegart: Done

src/components/CollateralChoice.tsx
```diff
@@ -35,7 +40,13 @@ const CollateralChoice = ({ id }: { id: string }) => {
...
   }
 
-  const isLoading = !(manager && metrics && params && price);
+  const isLoading = !(
+    manager &&
+    metrics &&
+    params &&
+    price &&
+    vaultFactoryParams
+  );
```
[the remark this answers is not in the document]
samsiegart: Yea I just didn't bother updating this one, done

src/utils/displayFunctions.ts
```diff
@@ -8,6 +8,11 @@ import type { BrandInfo } from 'store/app';
 import type { PriceDescription, Ratio } from 'store/vaults';
 import type { Brand, Amount } from '@agoric/ertp/src/types';
 
+// XXX: Kludge until we get a price authority with the same brands vault
+// manager uses https://github.com/Agoric/agoric-sdk/issues/6765.
```
turadg, line 12: 👍

src/components/NewVault.tsx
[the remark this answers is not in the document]
turadg: IMO it's okay to send arguments to a function that ignores them but I agree here it's not necessary given how little performance impact it would have.

turadg (approved):

Issue #13, opened 2022-12-01
As Vaults user in ‘Your Vaults’ view, I need to be able to easily create a new vault


otoole-brendan:
## What is the Problem Being Solved?

This view needs to include a 'Create New Vault' CTA button (this is the alternative vault creation pathway if user does not start from landing page)

## Description of the Design

Use Figma design for reference.

- There should be a prominent and clear ‘Create New Vault’ CTA button

## Security Considerations
Unknown
## Test Plan
Smoke Test after implementation

otoole-brendan:
sam confirmed done-ish

Issue #12, opened 2022-12-01
As Vaults user, I need a ‘Your Vaults’ Dashboard view where I can see all my active, closed, liquidating and liquidated vaults.


otoole-brendan:
## What is the Problem Being Solved?

We need to implement a 'Your Vaults' view after user successfully creates a vault

## Description of the Design

1) General page requirements:

- [ ] Clear 'Create Vault' CTA button
- [ ] Top bar navigation to other dapps
- [ ] top right connect wallet button/functionality
- [ ] Vault-specific text and graphics
- [ ] Clickable Inter logo which brings users to inter.trade
- [ ] clickable PSM which brings users to psm.inter.trade

2) Refer to Figma design for reference https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=1876%3A10456&t=I65DB2Ve9pmj26hG-0

Tile view works well I think. Should be visually appealing and to include:

- collateral logo (so easily identifiable)

- The most prominent 'XXX USD' value in line with the logo is net equity (Value of collateral in USD less outstanding IST) - we may need to label this so it's more clear.

- the collateral amount locked

- vault type and name

- Current Collateralization Ratio (dynamic value that updates in line with rounded collateral fluctuations)

- Liquidation Ratio

- Interest Rate

- IST Debt

- Collateral Locked ($USD Value)

Also, clicking tile should being user to ‘Manage/Adjust Vaults’ view - it should be obviously clickable when a user hovers over so this interaction is not missed.

## Security Considerations
Unknown
## Test Plan
Smoke Test after implementation

PR #30, opened 2023-01-13
Show vaults

samsiegart:
refs https://github.com/Agoric/dapp-inter/issues/13
refs https://github.com/Agoric/dapp-inter/issues/12

The loading states for the  "Manage Vaults" view could be handled a little smoother. Right now the vaults just pop up as each one loads individually, async. This can be improved on further when we can load *only* the user's vaults (https://github.com/Agoric/agoric-sdk/issues/6665), since a lot of this code will need to be rewritten then.

For now this just establishes some basic vault loading/rendering:
image

### DIFF
--- a/src/components/ManageVaults.tsx
+++ b/src/components/ManageVaults.tsx
@@ -1,0 +1,23 @@
+import { useVaultStore } from 'store/vaults';
+import VaultSummary from 'components/VaultSummary';
+import { useAtomValue } from 'jotai';
+import { displayFunctionsAtom } from 'store/app';
+
+const ManageVaults = () => {
+  const vaults = useVaultStore(state => state.vaults);
+  const displayFunctions = useAtomValue(displayFunctionsAtom);
+  const content = displayFunctions ? (
+    [...vaults.entries()].map(([key]) => <VaultSummary id={key} key={key} />)
+  ) : (
+    <p>Loading...</p>
+  );
+
+  return (
+    <>
+      <h2>Manage vaults:</h2>
+      <div className="flex gap-4 flex-wrap p-1">{content}</div>
+    </>
+  );
+};
+
+export default ManageVaults;
--- a/src/components/ManageVaults.tsx
+++ b/src/components/ManageVaults.tsx
@@ -7,7 +7,9 @@ const ManageVaults = () => {
   const vaults = useVaultStore(state => state.vaults);
   const displayFunctions = useAtomValue(displayFunctionsAtom);
   const content = displayFunctions ? (
-    [...vaults.entries()].map(([key]) => <VaultSummary id={key} key={key} />)
+    [...vaults.entries()].map(([key]) => (
+      <VaultSummary key={key} vaultKey={key} />
+    ))
   ) : (
     <p>Loading...</p>
   );
--- a/src/components/VaultSummary.tsx
+++ b/src/components/VaultSummary.tsx
@@ -1,0 +1,51 @@
+import { useMemo } from 'react';
+import { useVaultStore } from 'store/vaults';
+import { useAtomValue } from 'jotai';
+import { displayFunctionsAtom } from 'store/app';
+import type { VaultKey } from 'store/vaults';
+
+type Props = {
+  id: VaultKey;
+};
+
+const VaultSummary = ({ id }: Props) => {
+  const { vaults, errors } = useVaultStore(state => ({
+    vaults: state.vaults,
+    errors: state.vaultErrors,
+  }));
+  const vault = vaults.get(id);
+  const error = errors.get(id);
+  const displayFunctions = useAtomValue(displayFunctionsAtom);
+
+  return useMemo(() => {
+    assert(vault, `Cannot render summary for nonexistent vault ${id}`);
+    assert(
+      displayFunctions,
+      `Cannot render summary for vault ${id} - missing vbank asset info.`,
+    );
+    const { displayAmount, displayBrandPetname } = displayFunctions;
+
+    if (error) {
+      return (
+        <div className="p-4 pt-2 border border-black border-solid">
+          <h3>{id}</h3>
+          <p>Error: {error.toString()}</p>
+        </div>
+      );
+    }
+
+    // TODO: Calculate and display total debt correctly.
+    return (
+      <div className="p-4 pt-2 border border-black border-solid">
+        <h3>{id}</h3>
+        <p>Status: {vault.vaultState}</p>
+        <p>
+          Locked: {displayAmount(vault.locked)}{' '}
+          {displayBrandPetname(vault.locked.brand)}
+        </p>
+      </div>
+    );
+  }, [vault, error, id, displayFunctions]);
+};
+
+export default VaultSummary;
--- a/src/components/VaultSummary.tsx
+++ b/src/components/VaultSummary.tsx
@@ -5,30 +5,30 @@ import { displayFunctionsAtom } from 'store/app';
 import type { VaultKey } from 'store/vaults';
 
 type Props = {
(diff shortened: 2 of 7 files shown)


src/components/ManageVaults.tsx
```diff
@@ -0,0 +1,23 @@
+import { useVaultStore } from 'store/vaults';
+import VaultSummary from 'components/VaultSummary';
+import { useAtomValue } from 'jotai';
+import { displayFunctionsAtom } from 'store/app';
+
+const ManageVaults = () => {
+  const vaults = useVaultStore(state => state.vaults);
+  const displayFunctions = useAtomValue(displayFunctionsAtom);
```
turadg, line 8: a little confusing to me that displayFunctions is the condition for rendering. What exactly is "loading" when they're not available?

src/components/VaultSummary.tsx
```diff
@@ -0,0 +1,51 @@
+import { useMemo } from 'react';
+import { useVaultStore } from 'store/vaults';
+import { useAtomValue } from 'jotai';
+import { displayFunctionsAtom } from 'store/app';
+import type { VaultKey } from 'store/vaults';
+
+type Props = {
+  id: VaultKey;
```
turadg, line 8: IIUC, `VaultKey` is a composite key of [manager's id in the factory, vault's id within the manager]. 

Since that's the key of `vaults` and `errors` maps, shouldn't this be,
```suggestion
  key: VaultKet;
```
It overlaps with React's `key` property but not in any problematic way.

src/service/vaults.ts
```diff
@@ -110,6 +112,8 @@ type VaultFactoryParamsUpdate = ValuePossessor<{
 
 type AgoricInstancesUpdate = ValuePossessor<Array<[string, unknown]>>;
 
+type VaultUpdate = ValuePossessor<VaultInfoChainData>;
+
 export const watchVaultFactory = (netconfigUrl: string) => {
   let isStopped = false;
```
turadg, line 118: looks like nothing changes this yet? if so please leave a comment explaining why it should stay here.

src/views/Vaults.tsx
```diff
@@ -1,13 +1,20 @@
...
+import { useCallback, useEffect, useState } from 'react';
 import { watchVaultFactory } from 'service/vaults';
 import CollateralChoices from 'components/CollateralChoices';
+import ManageVaults from 'components/ManageVaults';
 import { useVaultStore } from 'store/vaults';
 
+const Modes = {
+  create: 0,
+  manage: 1,
+};
```
turadg, line 12: [TS's enum](https://www.typescriptlang.org/docs/handbook/enums.html) would give you some helper types and reverse mapping.
```suggestion
enum Mode = {
  Create = 0,
  Manage = 1,
};
```

src/views/Vaults.tsx
```diff
@@ -18,22 +25,73 @@ const Vaults = () => {
...
           <div>Loading collateral choices...</div>
         )}
       {vaultManagerIds && brandToInfo && <CollateralChoices />}
     </>
   );
+
+  const manage = <ManageVaults />;
+
+  const buttonText = (() => {
+    switch (mode) {
```
turadg, line 45: with the enum TS can check that each possible value has a case:
```suggestion
    switch (mode: Mode) {
```

src/views/Vaults.tsx
```diff
@@ -18,22 +25,73 @@ const Vaults = () => {
...
+      case Modes.create:
+        return 'Back to Manage Vaults';
+      case Modes.manage:
+        return 'Add New Vault';
+      default:
+        return 'Manage Vaults';
```
turadg, line 51: how can `mode` be other than the two enumerated values ?

src/views/Vaults.tsx
```diff
@@ -18,22 +25,73 @@ const Vaults = () => {
...
+  const content = (() => {
+    switch (mode) {
+      case Modes.create:
+        return create;
+      case Modes.manage:
+        return manage;
+      default:
+        return create;
+    }
+  })();
```
turadg, line 64: wouldn't an object map work here?
```suggestion
  const content = modeFns[mode]();
```

src/views/Vaults.tsx
```diff
@@ -18,22 +25,73 @@ const Vaults = () => {
...
+  const switchMode = useCallback(() => {
+    switch (mode) {
+      case Modes.create:
+        setMode(Modes.manage);
+        break;
+      case Modes.manage:
+        setMode(Modes.create);
+        break;
+      default:
+        setMode(Modes.manage);
```
turadg, line 75: again, why a default?

src/views/Vaults.tsx
```diff
@@ -18,22 +25,73 @@ const Vaults = () => {
...
+        return create;
+      case Modes.manage:
+        return manage;
+      default:
+        return create;
+    }
+  })();
+
+  const switchMode = useCallback(() => {
+    switch (mode) {
```
turadg, line 67: you might also consider a transition map to express this.
```
setMode(nextMode[mode]);
```

src/service/vaults.ts
samsiegart: `watchVaultFactory` returns a closure which sets `isStopped` to `true` and it's invoked when the vaults view is unmounted https://github.com/Agoric/dapp-inter/blob/main/src/views/Vaults.tsx#L14-L18. Currently we only have the vaults view but when we eventually have psm, liquidations, bld boost etc. that would happen when the user navigates away.

src/components/VaultSummary.tsx
samsiegart: The overlap is problematic, trying to access `key` as a prop returns undefined. Just changed to `vaultKey` since it's better than `id`

src/components/ManageVaults.tsx
samsiegart: `displayFunctions` is derived from `brandToInfo` which is fetched from `agoricNames.brands`. I guess it's not super clear just from it being nullable that it's something that we want to wait to load. I'm considering switching the types to something more like [react-query ](https://react-query-v3.tanstack.com/reference/useQuery#_top) where you have not just `data?` but `{ data, isLoading, isError}` etc.

src/views/Vaults.tsx
samsiegart: Thanks, switched to an enum

src/views/Vaults.tsx
samsiegart: Ack, but removed the switch instead

src/views/Vaults.tsx
samsiegart: It can't, I guess this default is unnecessary especially since we have static type checking

src/views/Vaults.tsx
samsiegart: Yes done

src/views/Vaults.tsx
samsiegart: Did something similar PTAL

src/components/VaultSummary.tsx
turadg: > trying to access key as a prop returns undefined

TIL!

src/components/ManageVaults.tsx
turadg: > something more like [react-query ](https://react-query-v3.tanstack.com/reference/useQuery#_top)where you have not just data? but { data, isLoading, isError} etc.

SGTM

src/service/vaults.ts
turadg: ah right, I didn't see that.

turadg (approved):

PR #33, opened 2023-01-19
feat: root app layout and styles

samsiegart:
image


index.html
```diff
@@ -1,5 +1,8 @@
 <!DOCTYPE html>
-<html lang="en">
+<html
+  lang="en"
+  class="h-full bg-gradient-to-b from-interYellow to-interOrange bg-no-repeat bg-fixed"
```
turadg, line 4: why class the `<html>` instead of `<body>`?

src/App.tsx
```diff
@@ -50,34 +57,26 @@ const App = () => {
   }, [setError, leader, netConfig, setLeader]);
 
   return (
-    <>
+    <div>
```
turadg, line 60: to make sure I understand, was this necessary?

src/components/ConnectWalletButton.tsx
```diff
@@ -20,24 +20,28 @@ const ConnectWalletButton = () => {
       return 'Connecting';
     } else if (chainConnection) {
       // TODO, add a way to call walletService.disconnect.
-      return 'Keplr Connected';
+      return `...${chainConnection.address.substring(
+        chainConnection.address.length - 7,
+      )} Connected`;
```
turadg, line 25: consider a helper,
```suggestion
      return `...${truncatedAddress(chainConnection)} Connected`;
```

src/styles/globals.css
```diff
@@ -10,5 +12,9 @@
   --color-alternative-bright: #fafafa;
   --color-primary: #c084fc;
   --color-primary-dark: #a855f7;
+  --color-mineShaft: #484848;
```
turadg, line 15: this name is so specific as to not be a variable. Please name it for a role (e.g. `text`) or an approximate color (like `gray`)

src/components/MainContentCard.tsx
```diff
@@ -0,0 +1,17 @@
+import type { PropsWithChildren, ReactNode } from 'react';
+
+type Props = PropsWithChildren<{ header?: ReactNode; subheader?: ReactNode }>;
+
+const MainContentCard = ({ children, header, subheader }: Props) => {
+  return (
+    <div className="w-full bg-gradient-to-br from-[#fffcf2] to-[#ffffff] rounded-t-3xl rounded-b-xl shadow-[0px_34px_50px_0px_#ff7a1a]">
```
turadg, line 7: I'm not familiar with this brackets in className syntax. Is some part of the build chain making classes on the fly?

tailwind.config.cjs
```diff
@@ -1,7 +1,7 @@
 /** @type {import('tailwindcss').Config} */
 module.exports = {
   mode: 'jit', // update this line
-  content: ['./src/**/*.{js,jsx,ts,tsx}', './public/index.html'],
+  content: ['./src/**/*.{js,jsx,ts,tsx}', './index.html'],
```
turadg, line 4: 👍

turadg (commented):
Gettin' spiffy

index.html
samsiegart: For the background color, I want it to affect the entire page even if there's no content. This way we don't have to artificially stretch the body out to fill the whole screen just to apply the background.

tailwind.config.cjs
samsiegart: Yea just to be clear, tailwind prunes all classes not used by any of these files. So the classes I put in index.html that weren't used elsewhere were being pruned without this change, pretty cool!

src/components/ConnectWalletButton.tsx
samsiegart: Done

src/styles/globals.css
turadg: I get why the Figma would use it, because they're designing a screen that doesn't have to be maintained. In code we should not name literal colors. If there were more time I'd say to work with the designers to understand how they _use_ "mine shaft" and ask them to give it a semantic name. But I won't block on this.

src/components/MainContentCard.tsx
turadg: got it https://tailwindcss.com/docs/adding-custom-styles#using-arbitrary-values

turadg (approved):

PR #38, opened 2023-01-28
Read user vaults

samsiegart:
refs https://github.com/Agoric/dapp-inter/issues/12


src/components/ManageVaults.tsx
```diff
@@ -1,25 +1,54 @@
...
+
+  if (vaults?.size === 0) {
+    return <EmptyView>You have not opened any vaults yet.</EmptyView>;
+  }
+
+  if (!(vaults && displayFunctions)) {
+    return <EmptyView isShimmering={true}>Loading your vaults...</EmptyView>;
+  }
+
+  const content = [...vaults.entries()].map(([offerId]) => (
```
turadg, line 47: consider,
```suggestion
  const content = [...vaults.keys()].map((offerId) => (
```

src/components/NewVault.tsx
```diff
@@ -49,10 +50,8 @@ const NewVault = ({ id }: Props) => {
   const proposeOffer = () => {
     assert(isReady);
     makeOpenVaultOffer(
-      collateralPurse.pursePetname,
-      COLLATERAL_TO_LOCK,
-      istPurse.pursePetname,
-      IST_TO_BORROW,
+      AmountMath.make(collateralPurse.brand, COLLATERAL_TO_LOCK),
```
turadg, line 53: 👍 structure

src/components/VaultSummary.tsx
```diff
@@ -2,50 +2,56 @@ import { useMemo } from 'react';
...
-import type { VaultKey } from 'store/vaults';
+import SkeletonVaultSummary from 'components/SkeletonVaultSummary';
 
 type Props = {
-  vaultKey: VaultKey;
+  offerId: string;
 };
 
-const VaultSummary = ({ vaultKey }: Props) => {
+const VaultSummary = ({ offerId }: Props) => {
```
turadg, line 11: this is the offerId in which the vault was acquired? would help to document that as there are offerIds for changing and closed vaults too.

but I wonder about changing this to key off that. seems like a lower level implementation detail of how we get the vault data. Can't the `VaultStore` continue to key by `vaultKey`?

src/components/VaultSummary.tsx
```diff
@@ -2,50 +2,56 @@ import { useMemo } from 'react';
...
+      return <SkeletonVaultSummary />;
+    }
+
+    assert(vault.locked, 'Vault must be loading still');
+
     // TODO: Calculate and display total debt correctly.
     return (
       <div className="p-4 pt-2 border border-black border-solid">
-        <h3>{vaultKey}</h3>
+        <h3>{offerId}</h3>
```
turadg, line 46: why is this what the user wants to see?

src/service/vaults.ts
```diff
@@ -92,6 +91,78 @@ const watchPriceFeeds = () => {
...
+
+  const watchNewVaults = async (
+    subscribers: Record<string, Record<string, string>>,
+  ) => {
+    if (!useVaultStore.getState().vaults) {
+      useVaultStore.setState({ vaults: new Map() });
+    }
+
+    Object.entries(subscribers).forEach(([offerId, subscribers]) => {
+      // XXX: If a third party contract returns a similar looking offer result,
```
turadg, line 132: good point! any offer result can have a `vault` key. to know it's really a vault we have to watch that key under `published.vaultFactory` which is authoritative.

src/service/vaults.ts
```diff
@@ -92,6 +91,78 @@ const watchPriceFeeds = () => {
...
+      // better way to filter for real vaults the user owns?
+      const vaultSubscriber = (subscribers as VaultSubscribers).vault;
+      if (!vaultSubscriber || watchedVaults.has(offerId)) {
+        return;
+      }
+
+      watchedVaults.add(offerId);
+      const managerId = (subscribers as VaultSubscribers).asset
+        .split('.')
+        .pop();
```
turadg, line 143: this string logic should be split out into a function to make obvious what it's doing. it might also change before release and it would be helpful to have isolated.

src/service/vaults.ts
```diff
@@ -294,41 +320,45 @@ export const watchVaultFactory = (netconfigUrl: string) => {
...
 export const makeOpenVaultOffer = async (
-  fundPursePetname: string,
-  toLock: bigint,
-  intoPursePetname: string,
-  toBorrow: bigint,
+  toLock: Amount<'nat'>,
+  toBorrow: Amount<'nat'>,
 ) => {
-  const INVITATION_METHOD = 'makeLoanInvitation';
+  const INVITATION_METHOD = 'makeVaultInvitation';
```
turadg, line 336: 👍

src/components/VaultSummary.tsx
samsiegart: I think it's useful to know we are looking at the offerId. When we implement adjust/close offers we'll need to pass in an offerId for the continuing invitation, so we'll need to store it regardless. The VaultKey thing was from when it didn't have access to an offerId and was showing everyone's vaults at once.

src/components/VaultSummary.tsx
samsiegart: It's not, just for debugging right now, final content is TODO

src/service/vaults.ts
samsiegart: It could still get a fake vault path (in that case it'd return no data and we'd show a vault summary card with an error state), or a vault path for a vault you don't own (that could be malicious and trick you into sending a bogus offer). I think it's worth revisiting adding boardIds to these publicSubscriber objects. Even if they can't always have boardIds, some offers (like opening vaults) are created with boardIds specified, and it would be useful to track those back to their offer results.

samsiegart: I might be forgetting the implementation details though, I guess the smart wallet derives the path string, and the contract returns the actual reference to the subscriber? In that case we could trust that the path isn't spoofable?

samsiegart: Hmm, unless another contract could get ahold of the subscriber since it is public...

src/components/VaultSummary.tsx
turadg: > When we implement adjust/close offers we'll need to pass in an offerId for the continuing invitation, so we'll need to store it regardless

Yes we'll have to store the mapping of vaultId with the offerId that created the continuing invitations, but the UI code should not be concerned with that. It's only concerned with what vault it's displaying and operating on. The underlying stores and services can manage turning that into whatever request format is necessary to accomplish the goal.

src/service/vaults.ts
turadg: > I might be forgetting the implementation details though, I guess the smart wallet derives the path string, and the contract returns the actual reference to the subscriber? In that case we could trust that the path isn't spoofable?

Good to think through this. The smart wallet gets the path string from the object, but it's not branded so there's no way to verify what object is returning the string. StorageNodes can only be made under a parent path, but one could make an object that acts like a StorageNode but returns an arbitrary string.

Fortunately we don't have any risk of this until MN-3.

> I think it's worth revisiting adding boardIds to these publicSubscriber objects.

Before Vaults Release?

src/components/ManageVaults.tsx
samsiegart: Done thanks

src/components/VaultSummary.tsx
samsiegart: Okay changed to `vaultKey` and added `createdByOfferId` as a separate property on the vault.

src/service/vaults.ts
samsiegart: Right no risk until MN-3, so no work to be done at the moment.

src/service/vaults.ts

turadg (approved):
🚀

Issue #35, opened 2023-01-27
dapp-inter isn't currently tolerant of being mounted in a subdirectory


arirubinstein:
## Describe the bug

dapp-inter isn't currently tolerant of being mounted in a subdirectory

## To Reproduce
1. goto dapp-inter directory
2. `yarn build`
3. `python -m http.server`
4. http://localhost:8000/dist
5. See error on router

## Expected behavior

The app should be able to load even if mounted at a subdirectory to support being served via ipfs, e.g., 

https://cloudflare-ipfs.com/ipfs/QmcHm8wo6M2TYB9QYhHB4JnMogGT5S3tht6qvBwgPWBh9X/index.html

samsiegart:
The way to solve this would be using a [hash router](https://reactrouter.com/en/main/router-components/hash-router). So instead of http://app.inter.trade/vaults it would be http://app.inter.trade#/vaults

samsiegart:
But as far as I know we don't plan on the user accessing the dapp from e.g. https://cloudflare-ipfs.com/ipfs/QmcHm8wo6M2TYB9QYhHB4JnMogGT5S3tht6qvBwgPWBh9X/index.html, is there an issue where this is documented?

arirubinstein:
We do have this req now per https://github.com/Agoric/agoric-sdk/issues/4350

PR #39, opened 2023-01-28
feat: use hashrouter to support ipfs

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/35


samsiegart:
@arirubinstein Is it possible to use a subdomain url for the github action to avoid the same-origin local storage problems? I already ran into them with the network config url.

turadg (approved):
slick

Issue #2, opened 2022-12-01
As Vaults user, I need a ‘Create New Vault’ view to create and configure my new vault


otoole-brendan:
## What is the Problem Being Solved?

We need to implement a 'Create New Vault' view and flow

## Description of the Design

When user clicks 'Create Vault' from landing page, or from dashboard, they navigate to 'Create vault' view:

- Step 1: Show user the vault types available for selection ( See Figma design for reference: https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=1876%3A12067&t=M1K9MKcyPF9GJnCO-0)
(action/intended behavior -> user selects vault type and their selection is made obvious by border)

Each vault type card needs to show distinguishing parameters:

- [ ] Vault Type Name (i.e. Atom-A)

- [ ] MCR

- [ ] Interest Rate

- [ ] IST Available 

- [ ] Liquidation Ratio

- Step 2: See https://github.com/Agoric/dapp-inter/issues/3

## Security Considerations
Unknown
## Test Plan
Smoke Test after implementation

otoole-brendan:
@samsiegart I've added more detail and linked Figma design for reference. There's likely more in this ticket than you originally estimated now if you are to handle the flows around these pop-up dialogs to direct user to create smart wallet, approve etc. - so you may need to re-estimate based on this.

Not having these pop-ups for every hand-off is likely fine for early MVP but we'll definitely need a couple by launch

otoole-brendan:
@samsiegart for launch this will only show the single vault type we support

PR #41, opened 2023-01-29
feat: use vbank branded pricequotes

samsiegart:
refs https://github.com/Agoric/dapp-inter/issues/2

### DIFF
--- a/src/service/vaults.ts
+++ b/src/service/vaults.ts
@@ -26,9 +26,9 @@ const watchPriceFeeds = () => {
   let isStopped = false;
   const { leader, importContext } = appStore.getState();
 
-  // Map of brands to any manager that publishes a price quote of that brand.
-  // If two managers have the same brand, their price quotes are the same, so
-  // we only need to watch one.
+  // Map of collateral brands to any manager that publishes a price quote of
+  // that brand. If two managers have the same brand, their price quotes are
+  // the same (denoted in IST), so we only need to watch one.
   const brandsToWatch = new Map<Brand, string>();
 
   const watchedBrands = new Set<Brand>();
--- a/src/service/vaults.ts
+++ b/src/service/vaults.ts
@@ -1,6 +1,5 @@
 import { fetchRPCAddr, fetchVstorageKeys } from 'utils/rpc';
 import {
-  PriceDescription,
   useVaultStore,
   VaultInfoChainData,
   VaultManager,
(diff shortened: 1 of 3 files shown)


src/store/vaults.ts
```diff
@@ -56,6 +56,9 @@ export type VaultKey = string;
 export const keyForVault = (managerId: string, vaultId: string) =>
   `${managerId}.${vaultId}` as VaultKey;
 
+// XXX: find a way to get this from zoe.
```
turadg, line 59: This is good for now. https://github.com/Agoric/agoric-sdk/pull/6884 will provide it. Consider an `UNTIL` comment.

src/service/vaults.ts
```diff
@@ -1,60 +1,46 @@
...
   let isStopped = false;
   const { leader, importContext } = appStore.getState();
 
-  const brandsToWatch = new Set<Brand>();
-  const watchedBrands = new Set<Brand>();
-  let brandToInfo: Map<Brand, BrandInfo> | null = null;
+  // Map of brands to any manager that publishes a price quote of that brand.
+  // If two managers have the same brand, their price quotes are the same, so
+  // we only need to watch one.
+  const brandsToWatch = new Map<Brand, string>();
```
turadg, line 33: price quotes are for a tuple, in:BrandA and out:BrandB. So I think we need the key to be those tuples. Those are safe to deduplicate but not a single brand.

src/utils/displayFunctions.ts
```diff
@@ -3,16 +3,15 @@ import {
   stringifyRatio,
   stringifyValue,
 } from '@agoric/ui-components';
-import { AssetKind } from '@agoric/ertp';
+import { AssetKind, AmountMath } from '@agoric/ertp';
+import {
+  floorMultiplyBy,
+  makeRatioFromAmounts,
+} from '@agoric/zoe/src/contractSupport';
```
turadg, line 10: this should be declared in package.json as a dependency. I guess we don't have the lint rule for that yet. 

https://github.com/Agoric/dapp-inter/pull/42 takes care of adding the dep.

src/utils/displayFunctions.ts
```diff
@@ -70,12 +69,22 @@ export const makeDisplayFunctions = (brandToInfo: Map<Brand, BrandInfo>) => {
...
+      brandIn,
+      10n ** BigInt(brandInDecimals),
     );
+
+    const brandOutAmountPerUnitOfBrandIn = floorMultiplyBy(
+      unitAmountOfBrandIn,
+      makeRatioFromAmounts(amountOut, amountIn),
+    );
+
+    return '$' + displayAmount(brandOutAmountPerUnitOfBrandIn);
```
turadg, line 87: this assumes `brandOut` is IST?

turadg (changes requested):

src/service/vaults.ts
samsiegart: Are we going to get vault price feeds where brand out is not IST? Tuples would be messy since we have to serialize them as strings.

src/utils/displayFunctions.ts
samsiegart: Yes, will it ever not be?

src/service/vaults.ts
turadg: > Are we going to get vault price feeds where brand out is not IST?

Yes, I believe that's on the roadmap. Though it's probably far enough out to disregard for me.

> Tuples would be messy since we have to serialize them as strings.

Ok, let's leave it as is. But please clarify that we're tracking _collateral_ brands and assuming the debt brand is IST.

src/utils/displayFunctions.ts
turadg: discussed in https://github.com/Agoric/dapp-inter/pull/41#discussion_r1090053754 . resolving this thread.

src/store/vaults.ts
samsiegart: Done

src/service/vaults.ts
samsiegart: Done

src/utils/displayFunctions.ts
samsiegart: Rebased

turadg (approved):

PR #42, opened 2023-01-29
refactor: marshal-contexts from smart-wallet

turadg:
It was moved to smart-wallet. Now it has types and the patch included.


samsiegart (approved):
🎉 Awesome more real types!

turadg:
> 🎉 Awesome more real types!

Slowly but surely!

PR #43, opened 2023-01-29
fix: restore import context patch

samsiegart:
Using patch-package still just to unblock and because I'm not sure if we want the version used by the smart wallet to be null-friendly.


turadg (approved):
Oops, I had forgot why patch was here. Maybe worth a comment in the patch code.

> not sure if we want the version used by the smart wallet to be null-friendly.

I'm not either so let's restore the status quo and revisit later.

PR #44, opened 2023-01-30
Styled collateral choices

samsiegart:
refs #2 

image


src/components/DebugCollateral.tsx
```diff
@@ -0,0 +1,138 @@
+import { useAtomValue } from 'jotai';
```
samsiegart, line 1: I just moved CollateralChoice.tsx to this file and added the comment. It's just handy to be able to refer to all the vault data.

samsiegart:
Will switch to main after the base PR goes in

turadg (approved):

PR #45, opened 2023-01-31
feat: vault creation layout and styles

samsiegart:
refs: #2 

image

### DIFF
--- a/src/components/AmountInput.tsx
+++ b/src/components/AmountInput.tsx
@@ -1,0 +1,36 @@
+import clsx from 'clsx';
+
+type InputProps = {
+  label?: string;
+  error?: string;
+  isDisabled?: boolean;
+};
+
+const AmountInput = ({ label, error, isDisabled = false }: InputProps) => {
+  return (
+    <div>
+      <div className="uppercase text-[#F4CCAE] text-xs leading-[19px] my-[2px]">
+        {label}
+      </div>
+      <div
+        className={clsx(
+          'w-fit transition rounded p-[2px] bg-gradient-to-r shadow-[0_12px_20px_-8px_#F0F0F0] focus-within:shadow-[0_12px_20px_-8px_rgba(255,83,0,0.2)]',
+          isDisabled
+            ? 'from-[#D9D9D9] to-[#A8A8A8]'
+            : 'from-[#FF7A1A] to-[#FFD81A]',
+        )}
+      >
+        <input
+          type="number"
+          placeholder="0.00"
+          min="0"
+          disabled={isDisabled}
+          className="p-2 bg-white rounded-sm outline-none text-sm font-medium w-64 placeholder:font-normal text-right"
+        ></input>
+      </div>
+      <div className="text-[#E22951] font-serif text-sm mt-2">{error}</div>
+    </div>
+  );
+};
+
+export default AmountInput;
--- a/src/components/AmountInput.tsx
+++ b/src/components/AmountInput.tsx
@@ -1,12 +1,12 @@
 import clsx from 'clsx';
 
-type InputProps = {
+type Props = {
   label?: string;
   error?: string;
-  isDisabled?: boolean;
+  disabled?: boolean;
 };
 
-const AmountInput = ({ label, error, isDisabled = false }: InputProps) => {
+const AmountInput = ({ label, error, disabled = false }: Props) => {
   return (
     <div>
       <div className="uppercase text-[#F4CCAE] text-xs leading-[19px] my-[2px]">
@@ -15,7 +15,7 @@ const AmountInput = ({ label, error, isDisabled = false }: InputProps) => {
       <div
         className={clsx(
           'w-fit transition rounded p-[2px] bg-gradient-to-r shadow-[0_12px_20px_-8px_#F0F0F0] focus-within:shadow-[0_12px_20px_-8px_rgba(255,83,0,0.2)]',
-          isDisabled
+          disabled
             ? 'from-[#D9D9D9] to-[#A8A8A8]'
             : 'from-[#FF7A1A] to-[#FFD81A]',
         )}
@@ -24,7 +24,7 @@ const AmountInput = ({ label, error, isDisabled = false }: InputProps) => {
           type="number"
           placeholder="0.00"
           min="0"
-          disabled={isDisabled}
+          disabled={disabled}
           className="p-2 bg-white rounded-sm outline-none text-sm font-medium w-64 placeholder:font-normal text-right"
         ></input>
       </div>
--- a/src/components/CollateralChoice.tsx
+++ b/src/components/CollateralChoice.tsx
@@ -71,10 +71,8 @@ const CollateralChoice = ({ id }: CollateralChoiceParams) => {
 
   const logoSrc = displayBrandIcon(metrics.totalCollateral.brand);
 
-  const collateralTitle =
-    displayBrandPetname(metrics.totalCollateral.brand).replace('Ibc', '') +
-    '-' +
-    id.substring('manager'.length);
+  // TODO: Come up with a naming scheme after rc0.
+  const collateralTitle = 'ATOM';
 
   return (
     <div className="w-fit px-6 pt-2 pb-4 bg-white rounded-[10.7px] shadow-[0_25px_35.6px_0_rgba(116,116,116,0.25)]">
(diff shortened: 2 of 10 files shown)


src/components/AmountInput.tsx
```diff
@@ -0,0 +1,36 @@
+import clsx from 'clsx';
+
+type InputProps = {
+  label?: string;
+  error?: string;
+  isDisabled?: boolean;
```
turadg, line 6: consider `disabled` like HTML input elements

src/components/AmountInput.tsx
```diff
@@ -0,0 +1,36 @@
+import clsx from 'clsx';
+
+type InputProps = {
```
turadg, line 3: not sure why this is called `InputProps` instead of merely `Props`. If it were `AmountInputProps` I'd understand it as being specific about the component. Though seems redundant in an `AmountInput` module with one other binding.

src/components/CollateralChoice.tsx
```diff
@@ -71,10 +71,8 @@ const CollateralChoice = ({ id }: CollateralChoiceParams) => {
 
   const logoSrc = displayBrandIcon(metrics.totalCollateral.brand);
 
-  const collateralTitle =
-    displayBrandPetname(metrics.totalCollateral.brand).replace('Ibc', '') +
-    '-' +
-    id.substring('manager'.length);
+  // TODO: Come up with a naming scheme after rc0.
```
turadg, line 74: might want to reference https://github.com/Agoric/agoric-sdk/issues/6126

src/components/ConfigureNewVault.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <h3 className="mb-3 font-serif font-bold leading-[26px]">Configure</h3>
+      <p className="font-serif text-[#666980] leading-[26px]">
+        Choose your vault parameters.
+      </p>
+      <div className="mt-12 flex gap-x-20 gap-y-6 flex-wrap">
+        <AmountInput
+          label="Atom to lock up *"
+          error="Need to obtain funds"
+        ></AmountInput>
+        <AmountInput label="Collateralization ratio *"></AmountInput>
```
turadg, line 15: AmountInput doesn't take `children`, right?
```suggestion
        <AmountInput label="Collateralization ratio *" />
```

src/components/CreateVault.tsx
```diff
@@ -1,44 +1,49 @@
...
   return (
     <>
-      <div className="flex justify-between mt-6">
+      <div className="flex justify-between mt-6 flex-wrap gap-4">
         <div className="font-serif font-medium text-2xl">
           Creating New Vault
         </div>
         <button
-          className="text-[#f9fafe] text-xs uppercase flex flex-row justify-center items-center p-3 bg-interPurple rounded-md shadow-[0_10px_14px_-4px_rgba(183,135,245,0.3)]"
+          className="transition mr-1 text-[#A3A5B9] uppercase rounded-[6px] border-2 border-solid border-[#A3A5B9] py-2 px-7 leading-[14px] font-semibold text-xs bg-[#A3A5B9] bg-opacity-0 hover:bg-opacity-10 active:bg-opacity-20"
```
turadg, line 24: why are the colors of this button bespoke? is it really different than all the other buttons will be?

I see it's also used in TableRow.

src/components/CreateVault.tsx
```diff
@@ -1,44 +1,49 @@
...
           onClick={buttonProps.onClick}
         >
           {buttonProps.text}
         </button>
       </div>
-      <div className="mt-16 grid grid-cols-11 gap-x-8">
-        <div className="col-span-7">
+      <div className="mt-16 grid grid-cols-11 lg:gap-x-[51px]">
```
turadg, line 30: choosing gaps that don't match the grid rhythm is suspect to me. do we really need 51px instead of an existing named value?

maybe because the gap below is 18.5px. maybe out of scope to address.

src/components/MainContentCard.tsx
```diff
@@ -1,15 +1,37 @@
...
+const TickerItem = ({ label, value }: TickerItemProps) => (
+  <div className="ml-16 h-12 leading-[48px]">
+    <span className="font-medium font-serif text-[#736D6D] mr-2">{label}</span>
+    <span className="font-bold font-serif text-mineShaft">{value}</span>
+  </div>
+);
+
+type Props = PropsWithChildren<{ header?: ReactNode }>;
+
+const MainContentCard = ({ children, header }: Props) => {
```
turadg, line 17: MainContentCard sounds like a style agnostic to the content, but this card always shows ticker items.

please restore the `subheader` prop or rename this card to describe the job it does.

src/components/NewVaultOfferSummary.tsx
```diff
@@ -0,0 +1,69 @@
...
+type TableRowProps = {
+  left: string;
+  right: string;
+};
+
+const TableRow = ({ left, right }: TableRowProps) => {
+  return (
+    <tr className="text-[13px] leading-[21px]">
+      <td className="text-[#A3A5B9]">{left}</td>
+      <td className="text-right text-[#666980] font-black">{right}</td>
```
turadg, line 12: this is used in three places. please name it.

src/components/NewVaultOfferSummary.tsx
```diff
@@ -0,0 +1,69 @@
...
+    </tr>
+  );
+};
+
+const NewVaultOfferSummary = () => {
+  const isDisabled = true;
```
turadg, line 18: I suppose this is because the UI can't make new vault offers yet.

src/views/Vaults.tsx
```diff
@@ -19,6 +19,10 @@ const PathDescription = ({ mode }: PathDescriptionProps) => {
     );
   }
 
+  if (mode === ViewMode.Manage) {
+    return <span className="text-[#9193A5]">Vaults</span>;
```
turadg, line 23: `#9193A5` is so close to `#A8A8A8` and `#A3A5B9` . do we need another gray?

src/components/AmountInput.tsx
samsiegart: Sure changed to just `Props`

src/components/AmountInput.tsx
samsiegart: Sure I guess it makes sense in this case to make it congruent with the input element, normally I like sticking to the naming convention. Done.

src/components/CollateralChoice.tsx
samsiegart: This isn't concerning purses but vault managers. We could have multiple vault managers necessitating a better naming scheme while still having a single purse per brand.

src/components/ConfigureNewVault.tsx
samsiegart: Yea done

src/components/MainContentCard.tsx
samsiegart: I removed `subheader` as a prop because it will always be the same.

Just changed to `MainContentWrapper` to clarify that it contains but is not comprised entirely of the main content.

src/components/NewVaultOfferSummary.tsx
samsiegart: Right, this will be determined by the state of the inputs eventually

src/components/CreateVault.tsx
samsiegart: For this and all other tailwind specific comments I filed this issue https://github.com/Agoric/dapp-inter/issues/46

I know some of these are reused but it won't be clear to me how many are actually reused, or what an appropriate way to name them is, until I get through all the components. And we don't have all the finalized designs yet even.

I also know some of them are very close to existing classes but right now I don't want to spend time with trial and error seeing if the changes are noticeable enough.

This is worth addressing but I think can be put off, it's a symptom of not being given a design system to work from but rather mockups.

src/components/CollateralChoice.tsx
turadg: oops I pasted the wrong link https://github.com/Agoric/agoric-sdk/issues/6518

src/components/MainContentCard.tsx
turadg: I would still expect `MainContent` anything to be about presenting some content (any content) as primary. This component specifies the brand subheading so I'd expect a name like `BrandsContentWrapper`.

not a blocker

turadg (approved):

PR #47, opened 2023-02-01
Vault Creation Flow

samsiegart:
refs #2 

Recording 2023-02-01 at 00 48 14

### DIFF
--- a/src/components/AmountInput.tsx
+++ b/src/components/AmountInput.tsx
@@ -1,35 +1,65 @@
-import clsx from 'clsx';
+import { useState } from 'react';
+import { useAtomValue } from 'jotai';
+import { displayFunctionsAtom } from 'store/app';
+import { stringifyValue, parseAsValue } from '@agoric/ui-components';
+import { AssetKind } from '@agoric/ertp';
+import type { Brand } from '@agoric/ertp/src/types';
+import StyledInput from './StyledInput';
 
 type Props = {
+  value?: bigint | null;
   label?: string;
   error?: string;
+  brand?: Brand | null;
+  onChange: (value: bigint) => void;
   disabled?: boolean;
 };
 
-const AmountInput = ({ label, error, disabled = false }: Props) => {
+const AmountInput = ({
+  value,
+  onChange,
+  brand,
+  label,
+  error,
+  disabled = false,
+}: Props) => {
+  const { getDecimalPlaces } = useAtomValue(displayFunctionsAtom) ?? {};
+
+  const decimalPlaces =
+    (brand && getDecimalPlaces && getDecimalPlaces(brand)) ?? 0;
+
+  const amountString = stringifyValue(value, AssetKind.NAT, decimalPlaces);
+
+  const [fieldString, setFieldString] = useState(
+    value === null ? '' : amountString,
+  );
+
+  const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = ev => {
+    const str = ev.target?.value?.replace('-', '').replace('e', '');
+    setFieldString(str);
+
+    const parsed = parseAsValue(str, AssetKind.NAT, decimalPlaces);
+    onChange(parsed);
+  };
+
+  const displayString =
+    value === parseAsValue(fieldString, AssetKind.NAT, decimalPlaces)
+      ? fieldString
+      : amountString;
+
   return (
-    <div>
-      <div className="uppercase text-[#F4CCAE] text-xs leading-[19px] my-[2px]">
-        {label}
-      </div>
-      <div
-        className={clsx(
-          'w-fit transition rounded p-[2px] bg-gradient-to-r shadow-[0_12px_20px_-8px_#F0F0F0] focus-within:shadow-[0_12px_20px_-8px_rgba(255,83,0,0.2)]',
-          disabled
-            ? 'from-[#D9D9D9] to-[#A8A8A8]'
-            : 'from-[#FF7A1A] to-[#FFD81A]',
-        )}
-      >
-        <input
-          type="number"
-          placeholder="0.00"
-          min="0"
-          disabled={disabled}
-          className="p-2 bg-white rounded-sm outline-none text-sm font-medium w-64 placeholder:font-normal text-right"
-        ></input>
-      </div>
-      <div className="text-[#E22951] font-serif text-sm mt-2">{error}</div>
-    </div>
+    <StyledInput
+      label={label}
+      error={error}
+      inputProps={{
+        type: 'number',
+        placeholder: '0.00',
+        min: '0',
+        disabled,
+        value: displayString,
+        onChange: handleInputChange,
+      }}
+    />
   );
 };
 
--- a/src/components/AmountInput.tsx
+++ b/src/components/AmountInput.tsx
@@ -35,6 +35,8 @@ const AmountInput = ({
   );
 
   const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = ev => {
+    // Inputs with type "number" allow these characters which don't apply to
+    // Amounts, just strip them.
     const str = ev.target?.value?.replace('-', '').replace('e', '');
     setFieldString(str);
 
@@ -42,6 +44,13 @@ const AmountInput = ({
     onChange(parsed);
   };
 
+  // Use the `fieldString` as an input buffer so the user can type values that
+  // would be overwritten by `stringifyValue`. For example, if the current
+  // input is "1.05", and you tried to change it to "1.25", on hitting
+  // backspace, `stringifyValue` would change it from "1.0" to "1.00",
+  // preventing you from ever editing it. Only let `amountString` override
+  // `fieldString` if the controlled input is trying to change it to a truly
+  // different value.
   const displayString =
     value === parseAsValue(fieldString, AssetKind.NAT, decimalPlaces)
       ? fieldString
--- a/src/components/CollateralChoice.tsx
+++ b/src/components/CollateralChoice.tsx
@@ -76,8 +76,9 @@ const CollateralChoice = ({ id }: CollateralChoiceParams) => {
   }
 
   if (!isReady) {
+    // TODO: Implement a better looking skeleton component.
     return (
-      <div className="w-fit p-4 pt-2 border border-solid border-black">
+      <div className="h-[248px] p-4 pt-2 border border-solid border-black">
         Loading...
       </div>
     );
--- a/src/components/CollateralChoice.tsx
+++ b/src/components/CollateralChoice.tsx
@@ -115,9 +115,16 @@ const CollateralChoice = ({ id, displayFunctions }: CollateralChoiceParams) => {
 
   const isSelected = selectedCollateralId === id;
 
+  const onClick = () => {
+    if (isSelected) {
+      setSelectedCollateralId(null);
+    } else {
+      setSelectedCollateralId(id);
+    }
+  };
   return (
     <button
-      onClick={() => setSelectedCollateralId(id)}
+      onClick={onClick}
       className={clsx(
         'w-fit px-6 pt-2 pb-4 bg-white rounded-[10.7px] cursor-pointer',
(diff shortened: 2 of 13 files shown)


src/components/CollateralChoice.tsx
```diff
@@ -75,8 +98,18 @@ const CollateralChoice = ({ id }: CollateralChoiceParams) => {
   // https://github.com/Agoric/agoric-sdk/issues/6518
   const collateralTitle = 'ATOM';
 
+  const isSelected = selectedCollateralId === id;
+
   return (
-    <div className="w-fit px-6 pt-2 pb-4 bg-white rounded-[10.7px] shadow-[0_25px_35.6px_0_rgba(116,116,116,0.25)]">
+    <button
```
turadg, line 104: [Buttons can have content](https://css-tricks.com/use-button-element/#aa-buttons-can-have-content) but the table doesn't seems to fit the [content categories of a button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#technical_summary).

What motivates this? I see there's an `onClick` handler so that would work on a `div` too.

src/components/ConfigureNewVault.tsx
```diff
@@ -1,19 +1,99 @@
...
+  const selectedMetrics =
+    selectedCollateralId && metrics?.has(selectedCollateralId)
+      ? metrics.get(selectedCollateralId)
+      : null;
+
+  const collateralBrand = selectedMetrics
+    ? selectedMetrics.retainedCollateral.brand
+    : null;
+
+  const hasPriceFeed = !!collateralBrand && prices.has(collateralBrand);
```
turadg, line 45: it looks like the `!!` is to ensure this is a boolean. if that's important, please annotate,
```suggestion
  const hasPriceFeed: boolean = !!collateralBrand && prices.has(collateralBrand);
```

also consider for readability,
```suggestion
  const hasPriceFeed: boolean = !!(collateralBrand && prices.has(collateralBrand));
```

src/components/ConfigureNewVault.tsx
```diff
@@ -1,19 +1,99 @@
...
+  const borrowedBrand = selectedMetrics
+    ? selectedMetrics.totalDebt.brand
+    : null;
+
+  const selectedParams =
+    selectedCollateralId && params?.has(selectedCollateralId)
+      ? params.get(selectedCollateralId)
+      : null;
+
+  const isInputReady = hasPriceFeed && !!selectedParams && !!selectedMetrics;
```
turadg, line 56: ditto re hasPriceFeed

src/components/NewVaultOfferSummary.tsx
```diff
@@ -14,8 +26,93 @@ const TableRow = ({ left, right }: TableRowProps) => {
...
+    displayPercent && selectedParams
+      ? `${displayPercent(selectedParams.liquidationMargin, 0)}%`
+      : '--';
+
+  const collateralizationRatioForDisplay =
+    displayPercent && collateralizationRatio
+      ? `${displayPercent(collateralizationRatio, 0)}%`
+      : '--';
+
+  const isDisabled =
```
turadg, line 102: all these bangs. ditto re hasPriceFeed for boolean handling and also this makes more sense as an `&&`. All these conditions must be met.

Above you name the state semantically like `isInputReady` and then map that to the UI concern of "disabled" in that property. I think that's a better way to go here too.

```
  const hasErrors = collateralizationRatioError || toLockError || toReceiveError;
  const canCreateVault: boolean = 
    !hasErrors &&
    selectedMetrics && 
    selectedParams && 
```

src/components/NewVaultOfferSummary.tsx
```diff
@@ -14,8 +26,93 @@ const TableRow = ({ left, right }: TableRowProps) => {
...
+    !!collateralizationRatioError ||
+    !!toLockError ||
+    !!toReceiveError ||
+    !selectedMetrics ||
+    !selectedParams ||
+    !factoryParams ||
+    !depositAmount ||
+    !borrowAmount ||
+    !offerSigner?.isDappApproved;
```
turadg, line 111: I think this one could return undefined.

src/utils/vaultMath.ts
```diff
@@ -0,0 +1,66 @@
+import {
+  floorDivideBy,
+  floorMultiplyBy,
+} from '@agoric/zoe/src/contractSupport';
+import { AmountMath } from '@agoric/ertp';
+import { Ratio } from 'store/vaults';
+
+export const computeToReceive = (
```
turadg, line 8: up for including some unit tests?

do you think these utils would be worthwhile to offer from `@agoric/inter-protocol/tools`?

src/utils/vaultMath.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+
+  return floorDivideBy(receiveMargin, priceRate).value;
+};
+
+/**
+ * Returns negative if left < right, 0 if left = right,
+ * positive if left < right.
+ */
+export const compareRatios = (left: Ratio, right: Ratio) => {
```
turadg, line 50: I'm surprised contractSupport doesn't have this. We do have `ratioGTE` but that's boolean. 

a ratio comparator definitely belongs in contractSupport. not a release blocker but please include a TODO here to move it.

src/utils/vaultMath.ts
```diff
@@ -0,0 +1,66 @@
...
+  assert(
+    left.numerator.brand === right.numerator.brand &&
+      left.denominator.brand === right.denominator.brand,
+    'Cannot compare ratios of different brands',
+  );
+
+  const leftNumerator = left.numerator.value;
+  const leftDenominator = left.denominator.value;
+  const rightNumerator = right.numerator.value;
+  const rightDenominator = right.denominator.value;
```
turadg, line 60: since you're making shorter names for readability, consider `rightNum` and `rightDen`

also consider `a` and `b` instead of `left` and `right` as that's the common way to describe a [compareFn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#parameters).

src/components/CollateralChoice.tsx
samsiegart: Hmm yea it needs to be clickable because you click this card to select your collateral choice. I initially did a div with a click handler but the linter told me divs are non-interactive. So I could either give it role="button", and implement the keyboard listeners myself, or just make it a button. We might just have to break the rules on this one.

src/components/ConfigureNewVault.tsx
samsiegart: Did the wrapping thing, but auto-formatter removes the redundant type annotation.

src/components/ConfigureNewVault.tsx
samsiegart: Done

src/components/NewVaultOfferSummary.tsx
samsiegart: Done

src/components/NewVaultOfferSummary.tsx
samsiegart: Fixed up as above

src/utils/vaultMath.ts
samsiegart: Removed

src/utils/vaultMath.ts
samsiegart: Oops I didn't know about that, that suits our needs perfectly fine.

src/utils/vaultMath.ts
samsiegart: I realized I also need to take into account the loan fee, which seems to shave off a percentage of your value locked, reducing your overall collateralization ratio. I'm not entirely sure how to calculate it though. It definitely makes sense to put these helpers in `inter-protocol` because then we could test creating loans with the amounts generated.

samsiegart: After discussion I think the loan fee incorporation is just a nice-to-have. It won't cause offer safety or vault creation to fail, but might affect overall UX, but I think that needs more thought from a product standpoint.

src/components/CollateralChoice.tsx
turadg: fair. please leave a comment with the rationale

src/components/ConfigureNewVault.tsx
turadg: hmm, a formatter shouldn't do that. is there an eslint fixer that is? I think we should disable whatever rule is erasing type annotations.

src/utils/vaultMath.ts
turadg: ok let's come back to it later

src/components/AmountInput.tsx
```diff
@@ -1,35 +1,65 @@
...
+    (brand && getDecimalPlaces && getDecimalPlaces(brand)) ?? 0;
+
+  const amountString = stringifyValue(value, AssetKind.NAT, decimalPlaces);
+
+  const [fieldString, setFieldString] = useState(
+    value === null ? '' : amountString,
+  );
+
+  const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = ev => {
+    const str = ev.target?.value?.replace('-', '').replace('e', '');
```
turadg, line 38: I meant to ask in my last review… why remove '-' and 'e'? some examples of what these string values would be might help the reader

src/components/AmountInput.tsx
```diff
@@ -1,35 +1,65 @@
...
   error?: string;
+  brand?: Brand | null;
+  onChange: (value: bigint) => void;
   disabled?: boolean;
 };
 
-const AmountInput = ({ label, error, disabled = false }: Props) => {
+const AmountInput = ({
+  value,
+  onChange,
```
turadg, line 20: looks like this is mixing functionality of controlled input and an uncontrolled one. `value` comes in but `fieldString` and thus `displayString` is local state.

reasonable to clean this up?

turadg (commented):
Some more questions I didn't notice in my first pass.

src/components/ConfigureNewVault.tsx
samsiegart: I guess it's `@typescript-eslint`:
```
$ yarn lint
yarn run v1.22.19
warning ../package.json: No license field
$ eslint src

/home/samsiegart/dapp-inter/src/components/ConfigureNewVault.tsx
  45:9  error  Type boolean trivially inferred from a boolean literal, remove type annotation  @typescript-eslint/no-inferrable-types

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
  ```
  
  Just disabled it in eslintrc

File: src/components/NewVaultOfferSummary.tsx [hunk 11]
```diff
@@ -14,8 +26,98 @@ const TableRow = ({ left, right }: TableRowProps) => {
...
+    displayPercent && selectedParams
+      ? `${displayPercent(selectedParams.liquidationMargin, 0)}%`
+      : '--';
+
+  const collateralizationRatioForDisplay =
+    displayPercent && collateralizationRatio
+      ? `${displayPercent(collateralizationRatio, 0)}%`
+      : '--';
+
+  const hasErrors = !!(
```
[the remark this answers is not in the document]
samsiegart: Removed

src/components/NewVaultOfferSummary.tsx
```diff
@@ -14,8 +26,98 @@ const TableRow = ({ left, right }: TableRowProps) => {
...
+    collateralizationRatioError ||
+    toLockError ||
+    toReceiveError
+  );
+
+  const canCreateVault = !!(
```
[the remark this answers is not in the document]
samsiegart: Removed

src/components/AmountInput.tsx
samsiegart: Added a comment

src/components/AmountInput.tsx
samsiegart: Added a comment, this one is kind of tricky, PTAL

turadg (approved):
Thanks for the documentation. 👍

PR #48, opened 2023-02-02
feat: layout and styles for manage vaults view

samsiegart:
refs #12 

Recording on simulated slow network to show loading transitions:
Recording 2023-02-02 at 03 40 44


src/service/vaults.ts
```diff
@@ -75,15 +75,17 @@ const watchPriceFeeds = () => {
...
-  asset: string;
   vault: string;
 };
 
 /**
  * Ex. `asset: 'published.vaultFactory.manager0'` -> `'manager0'`.
  */
 const getManagerIdFromSubscribers = (subscribers: VaultSubscribers) =>
-  subscribers.asset.split('.').pop();
+  subscribers.vault.split('.').find(node => node.includes('manager'));
```
turadg, line 85: makes sense. NTS to put something like this into inter-protocol/tools so we have a test that ensures it.

turadg (approved):
Thanks for the screen recording!

PR #49, opened 2023-02-02
fix: fix toast message

samsiegart:


turadg (approved):
feel free to slip something this small into other PRs

PR #50, opened 2023-02-03
Vault Creation Page Improvements

samsiegart:
refs #2 

image


src/components/VaultCreationDialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+};
+
+const VaultCreationDialog = ({ isOpen, onClose }: Props) => {
+  const walletUrl = useAtomValue(walletUiHrefAtom);
+  const setViewMode = useSetAtom(viewModeAtom);
+
+  const goToWallet = () => {
+    window.open(walletUrl, 'wallet');
```
turadg, line 17: consider using the third param to specify ideal window dimensions

src/utils/displayFunctions.ts
```diff
@@ -21,8 +21,10 @@ const getLogoForBrandPetname = (brandPetname: string) => {
   }
 };
 
-export const displayPetname = (pn: Array<string> | string) =>
-  Array.isArray(pn) ? pn.join('.') : pn;
+// XXX We remove the "Ibc" prefix so e.g. "IbcATOM" becomes "ATOM". Should
+// this be mapped out more explicitly?
```
turadg, line 25: good question. yeah I think it would be safer to do this only for known transformations. It could would also be more clear when searching the code.

tailwind.config.cjs
```diff
@@ -43,6 +43,23 @@ module.exports = {
           'Times',
           'serif',
         ],
+        altSans: [
```
turadg, line 46: surprised to see a second sans-serif font. to make clear that the fallback is our regular sans, please define this in terms of the other. e.g.
```
altSans: [
  'Mulish',
  ...sans,
]
```

src/components/CreateVault.tsx
```diff
@@ -107,7 +107,7 @@ const CreateVault = () => {
           Creating New Vault
         </div>
         <button
-          className="transition mr-1 text-[#A3A5B9] uppercase rounded-[6px] border-2 border-solid border-[#A3A5B9] py-2 px-7 leading-[14px] font-semibold text-xs bg-[#A3A5B9] bg-opacity-0 hover:bg-opacity-10 active:bg-opacity-20"
+          className="font-altSans transition mr-1 text-[#A3A5B9] uppercase rounded-[6px] border-2 border-solid border-[#A3A5B9] py-3 px-7 leading-[14px] font-bold text-xs bg-gray-500 bg-opacity-0 hover:bg-opacity-10 active:bg-opacity-20"
```
turadg, line 110: I see this new font applied to four buttons. They all share these features `font-bold font-altSans uppercase` (maybe more). Consider naming this button styling case with a classname.

turadg (approved):

src/components/VaultCreationDialog.tsx
samsiegart: I think it's preferable to just let it open in a new tab if possible.

src/components/CreateVault.tsx
samsiegart: Done

tailwind.config.cjs
samsiegart: Yea but it does look a lot nicer on the buttons for that small text. And done thanks!

src/utils/displayFunctions.ts
samsiegart: Done

PR #51, opened 2023-02-03
Vault summary

samsiegart:
refs: #12

### DIFF
--- a/src/components/MainContentWrapper.tsx
+++ b/src/components/MainContentWrapper.tsx
@@ -7,7 +7,6 @@ import { useAtomValue } from 'jotai';
 import { PropsWithChildren, ReactNode, useMemo } from 'react';
 import { displayFunctionsAtom } from 'store/app';
 import { useVaultStore } from 'store/vaults';
-import { addCommas } from 'utils/displayFunctions';
 
 type TickerItemProps = {
   label: string;
@@ -37,19 +36,20 @@ const MainContentWrapper = ({ children, header }: Props) => {
     let numVaultsForDisplay = '--';
 
     if (displayFunctions && metrics.size === managerIds?.length) {
-      let areAllPricesLoaded = true;
-      [...metrics.values()].forEach(() => {
-        const collateralBrand = metrics.get(managerIds[0])?.totalCollateral
-          .brand;
-        if (!(collateralBrand && prices.get(collateralBrand))) {
-          areAllPricesLoaded = false;
-        }
-      });
+      const areCollateralPricesLoaded = [...metrics.values()].every(m =>
+        prices.get(m.totalCollateral.brand),
+      );
 
-      if (areAllPricesLoaded) {
+      if (areCollateralPricesLoaded) {
         const { displayAmount } = displayFunctions;
 
-        const debtBrand = metrics.get(managerIds[0])?.totalDebt.brand;
+        const firstManagerMetrics = metrics.get(managerIds[0]);
+        assert(
+          firstManagerMetrics,
+          'Metrics arent loaded yet, cannot calculate info.',
+        );
+        // All managers use the same debt.
+        const debtBrand = firstManagerMetrics.totalDebt.brand;
 
         let totalDebt = AmountMath.makeEmpty(debtBrand);
         let totalLocked = AmountMath.makeEmpty(debtBrand);
@@ -69,9 +69,9 @@ const MainContentWrapper = ({ children, header }: Props) => {
           },
         );
 
-        totalDebtForDisplay = `$${addCommas(displayAmount(totalDebt, 0))}`;
-        tvlForDisplay = `$${addCommas(displayAmount(totalLocked, 0))}`;
-        numVaultsForDisplay = addCommas(numVaults.toString());
+        totalDebtForDisplay = `${displayAmount(totalDebt, 0, true, true)}`;
+        tvlForDisplay = `${displayAmount(totalLocked, 0, true, true)}`;
+        numVaultsForDisplay = new Intl.NumberFormat().format(numVaults);
       }
     }
 
--- a/src/components/MainContentWrapper.tsx
+++ b/src/components/MainContentWrapper.tsx
@@ -69,8 +69,8 @@ const MainContentWrapper = ({ children, header }: Props) => {
           },
         );
 
-        totalDebtForDisplay = `${displayAmount(totalDebt, 0, true, true)}`;
-        tvlForDisplay = `${displayAmount(totalLocked, 0, true, true)}`;
+        totalDebtForDisplay = `${displayAmount(totalDebt, 0, 'usd')}`;
+        tvlForDisplay = `${displayAmount(totalLocked, 0, 'usd')}`;
         numVaultsForDisplay = new Intl.NumberFormat().format(numVaults);
       }
     }
--- a/src/components/MainContentWrapper.tsx
+++ b/src/components/MainContentWrapper.tsx
@@ -1,4 +1,13 @@
-import type { PropsWithChildren, ReactNode } from 'react';
+import { AmountMath } from '@agoric/ertp';
+import {
+  ceilMultiplyBy,
+  makeRatioFromAmounts,
+} from '@agoric/zoe/src/contractSupport';
+import { useAtomValue } from 'jotai';
+import { PropsWithChildren, ReactNode, useMemo } from 'react';
+import { displayFunctionsAtom } from 'store/app';
+import { useVaultStore } from 'store/vaults';
+import { addCommas } from 'utils/displayFunctions';
 
 type TickerItemProps = {
   label: string;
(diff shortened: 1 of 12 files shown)


src/utils/displayFunctions.ts
```diff
@@ -21,6 +21,25 @@ const getLogoForBrandPetname = (brandPetname: string) => {
   }
 };
 
+export const addCommas = (stringifiedValue: string) => {
```
turadg, line 24: this is locale-specific and will be wrong in some countries. fortunately there are platform facilities for it. please use `Intl.NumberFormat` instead (within the display functions since it's part of the stringification)

src/views/Vaults.tsx
```diff
@@ -54,6 +55,9 @@ const Vaults = () => {
         <PathDescription mode={mode} />
       </div>
       <div className="text-red-600 text-lg mt-4">
+        {vaultFactoryInstanceHandleLoadingError && (
```
turadg, line 58: having empty divs if there's no error doesn't seem so bad.

yarn.lock
```diff
@@ -21,6 +21,11 @@
   resolved "https://registry.yarnpkg.com/@agoric/assert/-/assert-0.5.2-dev-669ad6d.0.tgz#91ea15ddd228d2209179103dae1337267baa07e8"
   integrity sha512-yUPwL6fqdsdtmVdQqw9s6+2O5HXt317GXdzKrMkzqJyEXNEPsFqFHRrsMJoVZZFje3Xi1r37dPVE6i2QSCwcUg==
 
+"@agoric/assert@0.5.2-dev-caf0bf0.0+caf0bf0":
```
turadg, line 24: are the `yarn.lock` changes necessary? I don't see any package.json changes

src/utils/displayFunctions.ts
samsiegart: We lose precision with `Number` but maybe that's okay for displaying prices because of practical limits. For amounts it could be a problem if we ever deal with an asset with a very large total supply, and I was just assuming it was okay to use a standard price format because we're dealing with USD. Not sure what's better.

samsiegart: Actually in practice it seems the precision issue is a lot easier to hit for decimals:
```
new Intl.NumberFormat().format(Number('1.232421'))
'1.232'
new Intl.NumberFormat().format(Number('1.232421234'))
'1.232'
new Intl.NumberFormat().format(Number('100000000011231.23242'))
'100,000,000,011,231.23'
new Intl.NumberFormat().format(Number('100000000000000023452542523452453253253'))
'100,000,000,000,000,020,000,000,000,000,000,000,000'
```

turadg: > standard price format because we're dealing with USD

USD is what the number is counting, but numbers themselves are locale-specific. For example, in Argentina it would be "$ 1.234,56".

> precision issue

Hmm. I hadn't considered that. Try https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toLocaleString

src/components/MainContentWrapper.tsx
```diff
@@ -15,13 +24,71 @@ const TickerItem = ({ label, value }: TickerItemProps) => (
...
+
+    if (displayFunctions && metrics.size === managerIds?.length) {
+      let areAllPricesLoaded = true;
+      [...metrics.values()].forEach(() => {
+        const collateralBrand = metrics.get(managerIds[0])?.totalCollateral
+          .brand;
+        if (!(collateralBrand && prices.get(collateralBrand))) {
+          areAllPricesLoaded = false;
+        }
+      });
```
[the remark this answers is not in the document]
samsiegart: Awesome thanks, done.

src/components/MainContentWrapper.tsx
```diff
@@ -15,13 +24,71 @@ const TickerItem = ({ label, value }: TickerItemProps) => (
...
+
+      if (areAllPricesLoaded) {
+        const { displayAmount } = displayFunctions;
+
+        const debtBrand = metrics.get(managerIds[0])?.totalDebt.brand;
```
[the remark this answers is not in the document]
samsiegart: Added an assert

src/components/MainContentWrapper.tsx
```diff
@@ -15,13 +24,71 @@ const TickerItem = ({ label, value }: TickerItemProps) => (
...
+
+        let totalDebt = AmountMath.makeEmpty(debtBrand);
+        let totalLocked = AmountMath.makeEmpty(debtBrand);
+        let numVaults = 0;
+
+        [...metrics.values()].forEach(
```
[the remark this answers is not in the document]
samsiegart: True, I wasn't sure if this would be easier to read. I don't think it's worth rewriting at this point though.

src/components/ManageVaults.tsx
```diff
@@ -58,7 +58,11 @@ const ManageVaults = () => {
     content = (
       <div className="mt-12 flex flex-wrap gap-x-6 gap-y-8 justify-center xl:justify-start xl:px-2">
         {[...vaults.keys()].map(vaultKey => (
-          <VaultSummary key={vaultKey} vaultKey={vaultKey} />
+          <VaultSummary
+            key={vaultKey}
+            vaultKey={vaultKey}
+            displayFunctions={displayFunctions}
```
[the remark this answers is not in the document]
samsiegart: It used to `assert` that `displayFunctions` wasn't null (because the value returned from the atom is nullable), and it was implicitly up to `VaultSummary`'s parent component to not render it if `displayFunctions` was null.  I made it so that the parent component must explicitly pass a non-null `displayFunctions` so that someone wouldn't accidently render a `VaultSummary` in the wrong place without checking for `displayFunctions` first (which could error non-deterministically depending on loading times).

Though I guess instead of asserting it could just show a loading state if it's null... hmm yea that's better, done.

src/components/VaultSummary.tsx
```diff
@@ -77,14 +111,45 @@ const VaultSummary = ({ vaultKey }: Props) => {
...
+
+    const totalLockedValue = ceilMultiplyBy(
+      locked,
+      makeRatioFromAmounts(price.amountOut, price.amountIn),
+    );
+
+    const maximumLockedValueForLiquidation = ceilMultiplyBy(
+      totalDebt,
+      params.liquidationMargin,
+    );
```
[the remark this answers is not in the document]
samsiegart: I'd like to defer until after I start the vault adjustment view. I'm not sure how many of these are bespoke and they're not too complex. The other functions are more critical for offer safety.

src/utils/displayFunctions.ts
samsiegart: I just figured it might not be too shocking to see USD in US locale, just not the best UX.

`BigInt.toLocaleString()` doesn't work because it doesn't know about decimal places. I realized the precision isn't really an issue for cases where we're showing 0 or 2 decimal places though, so just made it work for that case. PTAL

src/views/Vaults.tsx
samsiegart: True, done

src/components/VaultSummary.tsx
```diff
@@ -77,14 +111,45 @@ const VaultSummary = ({ vaultKey }: Props) => {
...
+    };
+
+    let netValueSignum;
+    let netVaultValue;
+    if (AmountMath.isGTE(totalLockedValue, totalDebt)) {
+      netVaultValue = AmountMath.subtract(totalLockedValue, totalDebt);
+    } else {
+      netVaultValue = AmountMath.subtract(totalDebt, totalLockedValue);
+      netValueSignum = '-';
+    }
```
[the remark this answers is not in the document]
samsiegart: Done

yarn.lock
samsiegart: I did add @agoric/inter-protocol as a dep

turadg: derp, I didn't see package.json in the list because it was at the top :)

src/components/MainContentWrapper.tsx
```diff
@@ -15,13 +23,72 @@ const TickerItem = ({ label, value }: TickerItemProps) => (
...
+      if (areCollateralPricesLoaded) {
+        const { displayAmount } = displayFunctions;
+
+        const firstManagerMetrics = metrics.get(managerIds[0]);
+        assert(
+          firstManagerMetrics,
+          'Metrics arent loaded yet, cannot calculate info.',
+        );
+        // All managers use the same debt.
+        const debtBrand = firstManagerMetrics.totalDebt.brand;
```
turadg, line 52: 👍

src/components/VaultSummary.tsx
```diff
@@ -53,22 +59,42 @@ type Props = {
...
-    assert(
-      displayFunctions,
-      `Cannot render summary for vault ${vaultKey} - missing vbank asset info.`,
-    );
-    const { displayBrandPetname, displayBrandIcon } = displayFunctions;
+  const metrics = vaultMetrics?.get(vault?.managerId ?? '');
+  const params = vaultGovernedParams?.get(vault?.managerId ?? '');
+  const brand = metrics?.totalCollateral?.brand;
+  const price = brand && prices.get(brand);
+  const manager = managers.get(vault?.managerId ?? '');
```
turadg, line 91: can that store have '' as a key?

src/service/vaults.ts
```diff
@@ -261,6 +261,7 @@ export const watchVaultFactory = (netconfigUrl: string) => {
       );
       assert(instanceEntry, 'Missing VaultFactory from agoricNames.instances');
       useVaultStore.setState({ vaultFactoryInstanceHandle: instanceEntry[1] });
+      return;
```
turadg, line 264: is this necessary?

src/utils/displayFunctions.ts
```diff
@@ -48,20 +66,45 @@ export const makeDisplayFunctions = (brandToInfo: Map<Brand, BrandInfo>) => {
     return stringifyRatio(ratio, getDecimalPlaces, placesToShow);
   };
 
-  const displayAmount = (amount: Amount, placesToShow?: number) => {
+  const displayAmount = (
+    amount: Amount,
+    placesToShow?: number,
+    localeFormat: boolean = false,
+    usdFormat: boolean = false,
```
turadg, line 73: multiple interdependent booleans calls for an enum.
```
format?: 'usd' | 'locale' | undefined
```

src/utils/displayFunctions.ts
turadg: lgtm. some feedback there but for this thread: please delete `addCommas`

src/service/vaults.ts
samsiegart: I figured it won't change, so better to not leave the follower open.

src/components/VaultSummary.tsx
samsiegart: Nope

src/utils/displayFunctions.ts
samsiegart: Nice, done

src/utils/displayFunctions.ts
samsiegart: Oops, accidently came back during rebase. removed

src/components/VaultSummary.tsx
turadg: so why pass it in to `get`?
```suggestion
  const manager = vault?.managerId && managers.get(vault.managerId);
```

src/service/vaults.ts
turadg: how does it change anything about the follower? `return` at the end of a function is a noop

turadg (approved):

src/service/vaults.ts
samsiegart: It's inside the for-loop, I think that should work? We'll revisit once we get the new casting improvements anyway though

src/components/VaultSummary.tsx
samsiegart: I just thought it was more succinct to cast to a string inside the arg

src/service/vaults.ts
turadg: shoot, reading from limited context in diff.

Issue #15, opened 2022-12-01
As Vaults user managing my vault, I need to be able to borrow additional IST or repay existing IST debt from said vault (‘Adjust Debt’)


otoole-brendan:
## What is the Problem Being Solved?

Being able to adjust vault debt by repaying or borrowing more IST is a core function of vault design

## Description of the Design

Should include below user interactions:

- User should be able to select ‘Borrow’ or ‘Repay’ from Action dropdown. Design show here https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=1876%3A12560&t=I65DB2Ve9pmj26hG-0

- User should be able to see their available IST balance to repay (or borrow more).

- User should be able to manually enter and/or max fill

## Security Considerations
Unknown
## Test Plan
Smoke Test after implementation

PR #52, opened 2023-02-05
feat: add adjust vault view with transition

samsiegart:
refs #15 

(The gif looks choppy, but you can see the animations for yourself in the ipfs deployments without a local chain running)
Recording 2023-02-05 at 11 56 52

### DIFF
--- a/src/views/Vaults.tsx
+++ b/src/views/Vaults.tsx
@@ -17,20 +17,24 @@ import type { VaultKey } from 'store/vaults';
 
 type PathDescriptionProps = { mode: ViewMode; adjustVaultKey: VaultKey | null };
 
+// Package doesn't export types. See:
+// https://github.com/jcoreio/react-view-slider.
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const ViewSlider = ReactViewSlider as unknown as FunctionComponent<any>;
 
 const PathDescription = ({ mode, adjustVaultKey }: PathDescriptionProps) => {
-  if (mode === ViewMode.Edit && adjustVaultKey) {
-    return (
-      <span>
-        <span className="text-[#9193A5]">Vaults&nbsp;&nbsp;/&nbsp;&nbsp;</span>
-        Adjusting Vault
-      </span>
-    );
-  }
-
   if (mode === ViewMode.Edit) {
+    if (adjustVaultKey) {
+      return (
+        <span>
+          <span className="text-[#9193A5]">
+            Vaults&nbsp;&nbsp;/&nbsp;&nbsp;
+          </span>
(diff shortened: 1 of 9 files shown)


src/views/Vaults.tsx
```diff
@@ -1,16 +1,36 @@
...
+import ReactViewSlider from 'react-view-slider';
+import type { VaultKey } from 'store/vaults';
 
-type PathDescriptionProps = { mode: ViewMode };
+type PathDescriptionProps = { mode: ViewMode; adjustVaultKey: VaultKey | null };
 
-const PathDescription = ({ mode }: PathDescriptionProps) => {
-  if (mode === ViewMode.Create) {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const ViewSlider = ReactViewSlider as unknown as FunctionComponent<any>;
```
samsiegart, line 21: This package doesn't export types so this is the only way I found to get ts to quiet down.

turadg: Ah. Consider explaining that in a code comment.

I looked at the package and it had Flow types so we could convert to TS and publish a DefinitelyTyped definition but prolly not worth it.

src/views/Vaults.tsx
```diff
@@ -1,16 +1,36 @@
...
+
+const PathDescription = ({ mode, adjustVaultKey }: PathDescriptionProps) => {
+  if (mode === ViewMode.Edit && adjustVaultKey) {
```
turadg, line 24: This shares the first condition with the next if. Please move the check for the second condition within that block.

src/vite-env.d.ts
```diff
@@ -46,3 +46,7 @@ declare module '@agoric/inter-protocol/src/interest-math' {
     currentCompoundedInterest: Ratio,
   ) => Amount<'nat'>;
 }
+
+declare module 'react-view-slider' {
+  export const ViewSlider;
```
turadg, line 51: I think if this described the type being exported you wouldn't need the ts-expect-error cast when you import it

turadg (approved):

src/vite-env.d.ts
samsiegart: Unfortunately it doesn't seem to be respecting it. I get `[JSX element type 'ViewSlider' does not have any construct or call signatures](https://github.com/s-yadav/react-number-format/issues/290#top)` unless I cast it here.

src/views/Vaults.tsx
samsiegart: Done

src/views/Vaults.tsx
samsiegart: Added a comment

Issue #14, opened 2022-12-01
As Vaults user managing my vault, I need to be able to deposit additional collateral or withdraw existing collateral (‘Adjust Collateral’) from said vault.


otoole-brendan:
## What is the Problem Being Solved?
Being able to adjust collateral by depositing more or withdrawing is a core feature of vaults design

## Description of the Design

Should include below interactions:

- User should be able to select ‘Deposit’ or ‘Withdraw’ from Action dropdown. See design here https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=1876%3A12560&t=I65DB2Ve9pmj26hG-0

- User should be able to see their available balance to add as further collateral.

- User should be able to manually enter and/or max fill

## Security Considerations
Unknown
## Test Plan
Smoke Test after implementation

PR #53, opened 2023-02-06
Adjust vault page layout

samsiegart:
refs: #14 

image

### DIFF
--- a/src/components/AdjustVault.tsx
+++ b/src/components/AdjustVault.tsx
@@ -6,7 +6,6 @@ import {
   ViewMode,
   viewModeAtom,
 } from 'store/vaults';
-import { AmountMath } from '@agoric/ertp';
 import {
   ceilMultiplyBy,
   makeRatioFromAmounts,
@@ -17,16 +16,24 @@ import clsx from 'clsx';
 import AdjustVaultForm from './AdjustVaultForm';
 import AdjustVaultSummary from './AdjustVaultSummary';
 import { useCallback } from 'react';
+import { netValue } from 'utils/vaultMath';
 
 const AdjustVault = () => {
   const vaultKey = useAtomValue(vaultKeyToAdjustAtom);
 
+  // We shouldn't ever see this component before display functions are loaded,
+  // so we don't need more graceful fallbacks. Just don't crash.
   const {
     displayPrice,
     displayPriceTimestamp,
     displayAmount,
     displayBrandPetname,
-  } = useAtomValue(displayFunctionsAtom) ?? {};
+  } = useAtomValue(displayFunctionsAtom) ?? {
+    displayPrice: () => '',
+    displayPriceTimestamp: () => '',
+    displayAmount: () => '',
+    displayBrandPetname: () => '',
+  };
 
   const { vaults, prices, managers } = useVaultStore(state => ({
     vaults: state.vaults,
@@ -48,8 +55,7 @@ const AdjustVault = () => {
     onClick: useCallback(() => setMode(ViewMode.Manage), [setMode]),
   };
 
-  const brand = locked.brand;
-  const price = brand && prices.get(brand);
+  const price = prices.get(locked.brand);
 
   const totalLockedValue =
     locked &&
@@ -64,12 +70,10 @@ const AdjustVault = () => {
     manager.compoundedInterest,
   );
 
-  const [netVaultValue, netValueSignum] = AmountMath.isGTE(
+  const [netVaultValue, isNetValueNegative] = netValue(
     totalLockedValue,
     totalDebt,
-  )
-    ? [AmountMath.subtract(totalLockedValue, totalDebt), undefined]
-    : [AmountMath.subtract(totalDebt, totalLockedValue), '-'];
+  );
 
   // TODO: Update dynamically.
   const vaultLabel = 'ATOM';
@@ -87,13 +91,13 @@ const AdjustVault = () => {
           <div>
             Current Price:{' '}
             <span className="text-[#00B1A6] font-medium text-lg">
-              {displayPrice && displayPrice(price)}
+              {displayPrice(price)}
             </span>
           </div>
           <div>
             Last Price Update:{' '}
             <span className="font-medium text-lg whitespace-nowrap">
-              {displayPriceTimestamp && displayPriceTimestamp(price)}
+              {displayPriceTimestamp(price)}
             </span>
           </div>
         </div>
@@ -104,10 +108,8 @@ const AdjustVault = () => {
             <VaultSymbol />
           </span>
           <span className="font-medium text-xl">
-            {vault.locked &&
-              displayAmount &&
-              displayAmount(vault.locked, 2, 'locale')}{' '}
-            {brand && displayBrandPetname && displayBrandPetname(brand)}
+            {displayAmount(locked, 2, 'locale')}{' '}
+            {displayBrandPetname(locked.brand)}
           </span>
         </div>
         <div className="text-lg">
@@ -115,23 +117,23 @@ const AdjustVault = () => {
           <span
             className={clsx(
               'font-medium',
-              netValueSignum ? 'text-red-500' : 'text-[#00B1A6]',
+              isNetValueNegative ? 'text-red-500' : 'text-[#00B1A6]',
             )}
           >
-            {netValueSignum}
-            {displayAmount && displayAmount(netVaultValue, 2, 'usd')}
+            {isNetValueNegative && '-'}
+            {displayAmount(netVaultValue, 2, 'usd')}
           </span>
         </div>
         <div className="text-lg">
           Collateral Value:{' '}
           <span className="font-medium">
-            {displayAmount && displayAmount(totalLockedValue, 2, 'usd')}
+            {displayAmount(totalLockedValue, 2, 'usd')}
           </span>
         </div>
         <div className="text-lg">
           Outstanding Debt:{' '}
           <span className="font-medium">
-            {displayAmount && displayAmount(totalDebt, 2, 'usd')}
+            {displayAmount(totalDebt, 2, 'usd')}
           </span>
         </div>
       </div>
--- a/src/components/AdjustVault.tsx
+++ b/src/components/AdjustVault.tsx
@@ -1,6 +1,11 @@
-import { useAtomValue } from 'jotai';
+import { useAtomValue, useSetAtom } from 'jotai';
 import { displayFunctionsAtom } from 'store/app';
-import { useVaultStore, vaultKeyToAdjustAtom } from 'store/vaults';
+import {
+  useVaultStore,
+  vaultKeyToAdjustAtom,
+  ViewMode,
+  viewModeAtom,
+} from 'store/vaults';
 import { AmountMath } from '@agoric/ertp';
 import {
   ceilMultiplyBy,
@@ -11,6 +16,7 @@ import VaultSymbol from 'svg/vault-symbol';
 import clsx from 'clsx';
 import AdjustVaultForm from './AdjustVaultForm';
 import AdjustVaultSummary from './AdjustVaultSummary';
+import { useCallback } from 'react';
 
 const AdjustVault = () => {
   const vaultKey = useAtomValue(vaultKeyToAdjustAtom);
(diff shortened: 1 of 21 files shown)


src/components/AdjustVaultForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+import StyledDropdown from './StyledDropdown';
+
+const AdjustVaultForm = () => {
+  return (
+    <div className="bg-white font-serif p-8 shadow-[0_40px_40px_-14px_rgba(116,116,116,0.25)] rounded-[20px] w-full">
+      <div className="font-bold mb-4">Adjust Collateral</div>
+      <div className="mb-12 grid grid-cols-2 gap-10">
+        <div className="text-[#9193A5] text-sm col-span-2 lg:col-span-1">
+          Deposit additional collateral or withdraw your existing collateral.
+          Select No Action to leave collateral unchanged.
```
turadg, line 12: "No Action" should be distinguished from the instructional text. Whatever style is used elsewhere for referring to UI options, or at least bold or italicize.

EDIT: I see it's put in quotes on line 38. That works too.

src/components/AdjustVaultForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </div>
+        <div className="col-span-2 lg:col-span-1">
+          <div className="mb-6">
+            <StyledDropdown
+              selection="No Action"
+              choices={['No Action']}
+              onSelection={() => {
+                console.log('Handle Action Selected');
```
turadg, line 20: low risk that you don't get to these but consider in the future making them more loud. E.g. logging as `console.error` and including TODO or FIXME.

src/components/AdjustVaultForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <div className="mb-12 grid grid-cols-2 gap-10">
+        <div className="text-[#9193A5] text-sm col-span-2 lg:col-span-1">
+          Borrow additional IST or repay your existing IST debt. Select “No
+          Action” to leave debt unchanged.
+        </div>
+        <div className="col-span-2 lg:col-span-1">
+          <div className="mb-6">
+            <StyledDropdown
+              selection="No Action"
+              choices={['No Action']}
```
turadg, line 45: this string appears repeatedly. I expect when you make a selection handler you'll want an enum

src/components/AdjustVaultSummary.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const { governedParams } = useVaultStore(state => ({
+    governedParams: state.vaultGovernedParams,
+  }));
+
+  const params = governedParams.get(managerId);
+
+  const collateralizationRatio = makeRatioFromAmounts(lockedValue, debt);
+
+  const { displayAmount, displayBrandPetname, displayPercent } =
+    useAtomValue(displayFunctionsAtom) ?? {};
```
turadg, line 64: if you do make defaults, they'd apply here too.

given the composable nature of Jotai, it could be another atom that returns existing `displayFuncions` if they're available and loading ones if they're not. Though now that I think about it why wouldn't we always want display functions to be available and usable? These UI components shouldn't have to condition on the state of that atom.

src/components/AdjustVaultSummary.tsx
```diff
@@ -0,0 +1,146 @@
...
+          <table className="w-full">
+            <tbody>
+              <TableRow left="Min. Collateralization Ratio" right="--" />
+              <TableRow
+                left="Liquidation Ratio"
+                right={`${
+                  params &&
+                  displayPercent &&
+                  displayPercent(params.liquidationMargin, 0)
+                }%`}
```
turadg, line 111: why doesn't `displayPercent` include the `%` char?

src/components/AmountInput.tsx
```diff
@@ -11,7 +11,7 @@ type Props = {
   label?: string;
   error?: string;
   brand?: Brand | null;
-  onChange: (value: bigint) => void;
+  onChange?: (value: bigint) => void;
```
turadg, line 14: consider giving it a default noop implementation instead, so you don't have to condition the call

src/components/CreateVault.tsx
```diff
@@ -72,16 +72,20 @@ const useVaultInputValidation = () => {
...
-      ({ brand }) => brand === selectedMetrics.totalCollateral.brand,
-    );
-
-    if (
-      !collateralPurse ||
-      (collateralPurse.currentAmount as Amount<'nat'>).value <
-        (valueToLock ?? 0n)
-    ) {
-      toLockError = 'Need to obtain funds';
+    if (!purses) {
```
turadg, line 75: I think our style is to avoid unnecessary negations. so,
```js
if (purses) {
  …
} else {
 toLockError = 'Need to connect wallet';
}
```

turadg: Looking more closely at `useVaultInputValidation` only the first handful of lines need to be in a hook. The rest of this logic would be better extracted into a function that could be tested as a unit. eg..
```js
const useVaultInputValidation = () => {
  return detectInputErrors({
    collateralizationRatio: useAtomValue(collateralizationRatioAtom),
    selectedCollateralId: useAtomValue(selectedCollateralIdAtom),
    valueToReceive: useAtomValue(valueToReceiveAtom),
    valueToLock: useAtomValue(valueToLockAtom),
    purses: useAtomValue(pursesAtom),
    vaultStoreState: useVaultStore.getState(),
  })
};
```

src/components/OfferSignerBridge.tsx
```diff
@@ -48,27 +49,22 @@ const WalletBridge = () => {
   const setIsDappApproved = useSetAtom(setIsDappApprovedAtom);
   const setOfferSigner = useSetAtom(offerSignerAtom);
   const chainConnection = useAtomValue(chainConnectionAtom);
-  const warningToastId = useRef<Id | null>(null);
-  const connectionSuccessfulToastId = useRef<Id | null>(null);
```
turadg, line 52: 👍

src/utils/displayFunctions.ts
```diff
@@ -111,7 +111,10 @@ export const makeDisplayFunctions = (brandToInfo: Map<Brand, BrandInfo>) => {
   };
 
   const displayPriceTimestamp = (price: PriceDescription) => {
-    return new Date(Number(price.timestamp) * 1000).toUTCString();
+    return new Intl.DateTimeFormat(navigator.language, {
+      timeStyle: 'medium',
+      dateStyle: 'short',
+    }).format(new Date(Number(price.timestamp) * 1000));
```
turadg, line 117: glad to see the formatting. what timezone is `timestamp` in? what assumptions about timezone does `new Date` make?

src/components/AdjustVault.tsx
```diff
@@ -1,5 +1,164 @@
...
+    displayBrandPetname,
+  } = useAtomValue(displayFunctionsAtom) ?? {};
+
+  const { vaults, prices, managers } = useVaultStore(state => ({
+    vaults: state.vaults,
+    prices: state.prices,
+    managers: state.vaultManagers,
+  }));
+
+  const vault = vaultKey && vaults?.get(vaultKey);
```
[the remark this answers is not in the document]
samsiegart: The key has to be a string and vaultKey is nullable

src/components/AdjustVault.tsx
```diff
@@ -1,5 +1,164 @@
...
+
+  const setMode = useSetAtom(viewModeAtom);
+
+  const backButtonProps = {
+    text: 'Back to vaults',
+    onClick: useCallback(() => setMode(ViewMode.Manage), [setMode]),
+  };
+
+  const brand = locked.brand;
+  const price = brand && prices.get(brand);
```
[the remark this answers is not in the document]
samsiegart: Done

src/components/AdjustVault.tsx
```diff
@@ -1,5 +1,164 @@
...
+    debtSnapshot.interest,
+    manager.compoundedInterest,
+  );
+
+  const [netVaultValue, netValueSignum] = AmountMath.isGTE(
+    totalLockedValue,
+    totalDebt,
+  )
+    ? [AmountMath.subtract(totalLockedValue, totalDebt), undefined]
+    : [AmountMath.subtract(totalDebt, totalLockedValue), '-'];
```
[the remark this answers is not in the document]
samsiegart: Done

src/components/AdjustVault.tsx
```diff
@@ -1,5 +1,164 @@
...
+          <div className="font-medium text-2xl">{vaultLabel}</div>
+          <div className="text-[#A3A5B9] text-sm">
+            #{vault.indexWithinManager}
+          </div>
+        </div>
+        <div className="flex gap-8">
+          <div>
+            Current Price:{' '}
+            <span className="text-[#00B1A6] font-medium text-lg">
+              {displayPrice && displayPrice(price)}
```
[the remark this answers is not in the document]
samsiegart: Done.

src/components/AdjustVault.tsx
```diff
@@ -1,5 +1,164 @@
...
+        </div>
+        <div className="text-lg">
+          Net Equity:{' '}
+          <span
+            className={clsx(
+              'font-medium',
+              netValueSignum ? 'text-red-500' : 'text-[#00B1A6]',
+            )}
+          >
+            {netValueSignum}
```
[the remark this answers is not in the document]
samsiegart: I think I'll hold off on that to keep `displayAmount` simpler

src/components/AdjustVaultForm.tsx
samsiegart: Done. In my defense I copy pasted both from figma, not sure why one didn't have quotes 😄

src/components/AdjustVaultForm.tsx
samsiegart: Added simple TODO, yea this is up next.

src/components/AdjustVaultForm.tsx
samsiegart: Yea

src/components/AdjustVaultSummary.tsx
samsiegart: Yes we want to show loading states on certain components depending on if we've loaded display functions or not (among other data). I'd like to maintain the nullability of `displayFunctionsAtom` for that purpose. For these components we just don't care about showing a loading state because we know we'll never get here without having loaded a vault summary card first, which requires display functions.

src/components/AdjustVaultSummary.tsx
samsiegart: When it's used to turn a ratio into a value for a number input it needs to not have a % char.

src/components/AmountInput.tsx
samsiegart: Done

src/utils/displayFunctions.ts
samsiegart: From my understanding timestamp we get from the chain is epoch time, so no timezone offset, but denoted in seconds rather than milliseconds. And the `Date` constructor expects epoch time in milliseconds.

src/components/CreateVault.tsx
samsiegart: I should've just made this a computed atom rather than hook in the first place, so just rewrote it as such. The atoms are pretty testable as well, probably even more so since we wouldn't have to map the args manually in the test.

samsiegart (commented):
PTAL

src/components/AdjustVaultSummary.tsx
```diff
@@ -60,8 +60,15 @@ const AdjustVaultSummary = ({
 
   const collateralizationRatio = makeRatioFromAmounts(lockedValue, debt);
 
-  const { displayAmount, displayBrandPetname, displayPercent } =
-    useAtomValue(displayFunctionsAtom) ?? {};
+  // We shouldn't ever see this component before display functions are loaded,
+  // so we don't need more graceful fallbacks. Just don't crash.
```
turadg, line 64: thx for explaining

src/store/createVault.ts
```diff
@@ -173,3 +176,78 @@ export const selectedCollateralIdAtom = atom(
     }
   },
 );
+
+export const inputErrorsAtom = atom<VaultCreationErrors>(get => {
```
turadg, line 180: nice

src/utils/vaultMath.ts
```diff
@@ -42,3 +43,8 @@ export const computeToLock = (
 
   return floorDivideBy(receiveMargin, priceRate).value;
 };
+
```
turadg, line 46: the return value could use some explaining
```suggestion
/**
 * @returns tuple of [value of difference, boolean of whether it's negative
 */
```

turadg (approved):
tidy :)

PR #58, opened 2023-02-08
fix: crash on opening vault adjustment page

samsiegart:


src/store/adjustVault.ts
```diff
@@ -12,9 +12,9 @@ import { pursesAtom } from './app';
 import { vaultKeyToAdjustAtom, vaultStoreAtom } from './vaults';
 
 export const vaultToAdjustAtom = atom(get => {
-  const key = get(vaultKeyToAdjustAtom);
   const { vaults, vaultManagers, prices, vaultGovernedParams, vaultMetrics } =
```
samsiegart, line 15: For some reason, this atom would do a double-update when first read, the first time `vaults` would be `null`, and right away it would update with its real value. However, the adjust vaults view would crash on first seeing a null `vaultToAdjust`. I think this is a bug in `jotai-zustand`, and had something to do with the above `useVaultStore(state => state.vaults)` causing it to produce a stale value. Moving to the atom in the above component fixed it.

samsiegart: And notably, this only would happen if you connected your wallet AFTER navigating to the manage vaults view

turadg (approved):
Thanks for the explanation

PR #60, opened 2023-02-09
feat: close vault dialog

### LINKED ISSUE #18: As Vaults user managing my vault, I need to be able to quickly and easily close my vault (repay all IST and withdraw collateral).

Our UX should ensure users can relatively easily close their vaults should they choose to (but not without a confirming such)

## Description of the Design

See Figma design for reference.

- User should be able to see prominent ‘Close Out Vault’ button. Shown here: https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=1876%3A12560&t=I65DB2Ve9pmj26hG-0

- Clicking should present confirmation dialog ‘Are you sure you want to close your vault? Shown here: https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=2003%3A10123&t=I65DB2Ve9pmj26hG-0

- It should clearly state ‘IST Debt to Pay Back’ amount and ‘Collateral to Receive’ amount.

- User should have choice to ‘Cancel’ and ‘Close Out Vault Type Name’ 

## Security Considerations
Unknown
## Test Plan
Smoke Test after implementation

samsiegart:
fixes #18 
Recording 2023-02-09 at 11 16 32

### DIFF
--- a/src/components/AdjustVaultForm.tsx
+++ b/src/components/AdjustVaultForm.tsx
@@ -54,7 +54,7 @@ const AdjustVaultForm = () => {
         <div className="mb-12 grid grid-cols-2 gap-10">
           <div className="text-[#9193A5] text-sm col-span-2 lg:col-span-1">
             Deposit additional collateral or withdraw your existing collateral.
-            Select &quot;No Action&quot; to leave collateral unchanged.
+            Select “No Action” to leave collateral unchanged.
           </div>
           <div className="col-span-2 lg:col-span-1">
             <div className="mb-6">
@@ -90,8 +90,8 @@ const AdjustVaultForm = () => {
         <div className="mt-8 font-bold mb-4">Adjust Debt</div>
         <div className="mb-12 grid grid-cols-2 gap-10">
           <div className="text-[#9193A5] text-sm col-span-2 lg:col-span-1">
-            Borrow additional IST or repay your existing IST debt. Select
-            &quot;No Action&quot; to leave debt unchanged.
+            Borrow additional IST or repay your existing IST debt. Select “No
+            Action” to leave debt unchanged.
           </div>
           <div className="col-span-2 lg:col-span-1">
             <div className="mb-6">
--- a/src/components/AdjustVaultForm.tsx
+++ b/src/components/AdjustVaultForm.tsx
@@ -47,6 +47,8 @@ const AdjustVaultForm = () => {
 
   const [isCloseVaultDialogOpen, setIsCloseVaultDialogOpen] = useState(false);
 
+  const isActive = vaultToAdjust?.vaultState === 'active';
+
   return (
     <>
       <div className="bg-white font-serif p-8 shadow-[0_40px_40px_-14px_rgba(116,116,116,0.25)] rounded-[20px] w-full">
@@ -119,15 +121,17 @@ const AdjustVaultForm = () => {
             />
           </div>
         </div>
-        <button
-          onClick={() => setIsCloseVaultDialogOpen(true)}
-          className="text-btn-xs flex items-center gap-3 transition text-[#E22951] rounded-[6px] border-2 border-solid border-[#E22951] h-8 px-4 leading-[14px] font-bold text-xs bg-[#E22951] bg-opacity-0 hover:bg-opacity-10 active:bg-opacity-20"
-        >
-          <span className="fill-[#E22951]">
-            <ErrorWarning />
-          </span>
-          Close Out Vault
-        </button>
+        {isActive && (
+          <button
+            onClick={() => setIsCloseVaultDialogOpen(true)}
+            className="text-btn-xs flex items-center gap-3 transition text-[#E22951] rounded-[6px] border-2 border-solid border-[#E22951] h-8 px-4 leading-[14px] font-bold text-xs bg-[#E22951] bg-opacity-0 hover:bg-opacity-10 active:bg-opacity-20"
+          >
+            <span className="fill-[#E22951]">
+              <ErrorWarning />
+            </span>
+            Close Out Vault
+          </button>
+        )}
       </div>
       <CloseVaultDialog
         isOpen={isCloseVaultDialogOpen}
--- a/src/components/AdjustVaultForm.tsx
+++ b/src/components/AdjustVaultForm.tsx
@@ -14,6 +14,8 @@ import ErrorWarning from 'svg/error-warning';
 import AmountInput from './AmountInput';
 import StyledDropdown from './StyledDropdown';
 import { PursesJSONState } from '@agoric/wallet-backend';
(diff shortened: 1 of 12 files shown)


src/components/AdjustVaultForm.tsx
```diff
@@ -43,84 +45,98 @@ const AdjustVaultForm = () => {
...
-              ]}
-              onSelection={setCollateralAction}
-              label="Action"
+    <>
+      <div className="bg-white font-serif p-8 shadow-[0_40px_40px_-14px_rgba(116,116,116,0.25)] rounded-[20px] w-full">
+        <div className="font-bold mb-4">Adjust Collateral</div>
+        <div className="mb-12 grid grid-cols-2 gap-10">
+          <div className="text-[#9193A5] text-sm col-span-2 lg:col-span-1">
+            Deposit additional collateral or withdraw your existing collateral.
+            Select &quot;No Action&quot; to leave collateral unchanged.
```
turadg, line 57: consider Unicode quote chars for readability,
```suggestion
            Select ‘No Action’ to leave collateral unchanged.
```

src/components/CloseVaultDialog.tsx
```diff
@@ -0,0 +1,150 @@
...
+};
+
+const noticeProps = {
+  initial: { height: 0, opacity: 0 },
+  animate: { height: 'auto', opacity: 1 },
+  transition: { type: 'tween' },
+};
+
+const message =
+  'Closing a vault is permanent and cannot be undone. To close your vault, please approve the offer in your wallet.';
```
turadg, line 38: "please" seems superfluous. It won't close unless they approve. This is something they're trying to accomplish so no need to request they do something
```suggestion
  'Closing a vault is permanent and cannot be undone. To close your vault, approve the offer in your wallet.';
```

src/components/CloseVaultDialog.tsx
```diff
@@ -0,0 +1,150 @@
...
+}: Props) => {
+  if (isOpen) {
+    assert(vaultOfferId, 'Need a vault to close');
+  }
+
+  const displayFunctions = useAtomValue(displayFunctionsAtom);
+  const { displayAmount, displayBrandPetname } = displayFunctions ?? {
+    displayAmount: () => '',
+    displayBrandPetname: () => '',
+  };
```
turadg, line 55: shouldn't `displayFunctions` be defined at this point?
```suggestion
  const { displayAmount, displayBrandPetname } = NonNullish(displayFunctions);
```

```js
import { NonNullish } from '@agoric/assert';
```

src/components/VaultAdjustmentDialog.tsx
```diff
@@ -25,10 +25,12 @@ const VaultAdjustmentDialog = ({ isOpen, onClose }: Props) => {
     setViewMode(ViewMode.Manage);
   };
 
+  const body = <p>{message}</p>;
+
   return (
     <BaseDialog
       title="Success: Offer Submitted"
-      message={message}
+      body={body}
```
turadg, line 33: consider inlining, unless that doesn't work for some reasoN?
```suggestion
      body={<p>{message}</p>}
```

src/components/VaultSummary.tsx
```diff
@@ -156,7 +157,9 @@ const VaultSummary = ({ vaultKey }: Props) => {
     );
 
     const maximumLockedPriceForLiquidation = {
-      amountIn: locked,
+      amountIn: AmountMath.isEmpty(locked)
+        ? AmountMath.make(locked.brand, 1n)
```
turadg, line 161: why is this necessary? please comment

src/store/adjustVault.ts
```diff
@@ -41,6 +41,14 @@ export const vaultToAdjustAtom = atom(get => {
     manager.compoundedInterest,
   );
 
+  // Prevent divide-by-zero if no debt.
+  const collateralizationRatio = AmountMath.isEmpty(totalDebt)
+    ? makeRatioFromAmounts(
+        totalLockedValue,
+        AmountMath.make(totalDebt.brand, 1n),
+      )
+    : makeRatioFromAmounts(totalLockedValue, totalDebt);
```
turadg, line 50:
```suggestion
  const nonzeroDebt = AmountMath.max(totalDebt, AmountMath.make(totalDebt.brand, 1n));
```
Then the line below could stay with a small change,
```js
collateralizationRatio: makeRatioFromAmounts(totalLockedValue, nonzeroDebt),
```
But… why have a collaterialization ratio at all if there's no debt? I would expect `undefined` or `null` since it doesn't make any sense.

src/components/VaultAdjustmentDialog.tsx
samsiegart: Nope that works I just wasn't sure if that was too messy. Done.

src/components/VaultSummary.tsx
samsiegart: Done

src/store/adjustVault.ts
samsiegart: Done.

Hmm yea this is just a safe way of handling it, what would we show instead? ∞?

turadg: Maybe safe from runtime failure but these seems less safe about what we're communicating to the user.

∞ would be better than what's here. But I think if there's no debt then the collateralization ratio is not applicable, so "n/a". I'm happy to leave it to Product to figure out what to display but I do request that we don't make up denominators in order to avoid zero division.

samsiegart: Yea just changed to make these undefined in that case and show "N/A". Seems like ts was having trouble inferring the types, thinking they were 'any', so I made them explicit so components would know they could be undefined.

samsiegart: Actually seems like it was having trouble inferring the return type for `Ratio`, just changed it to be more explicit.

src/store/adjustVault.ts
```diff
@@ -41,6 +59,10 @@ export const vaultToAdjustAtom = atom(get => {
     manager.compoundedInterest,
   );
 
+  const collateralizationRatio = AmountMath.isEmpty(totalDebt)
```
turadg, line 62: 👍 using ternary instead of `&&` for this

turadg (approved):

src/store/adjustVault.ts
```diff
@@ -26,6 +26,7 @@ type VaultToAdjust = {
   metrics: VaultMetrics;
   collateralizationRatio?: Ratio;
   createdByOfferId: string;
+  vaultState?: string;
```
turadg, line 29: I think this can be tightened to the list of possible strings

samsiegart: Done

PR #62, opened 2023-02-10
Update lint-test.yml

samsiegart:
We had this disabled

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -23,6 +23,8 @@ jobs:
       uses: actions/setup-node@v3
       with:
         node-version: ${{ matrix.node-version }}
+    - name: yarn install
+      run: yarn install
     - name: yarn build
       run: yarn build
     - run: yarn lint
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-test.yml
```diff
@@ -7,9 +7,6 @@ on:
   push:
     branches: [main]
   pull_request:
-  repository_dispatch:
-    types: [test-agoric-sdk]
-  # FIXME: Deprecated build from schedule.
   schedule:
     - cron: "0 12 * * *"
```
turadg, line 11: does this run a nightly build/test even if nothing has changed?

turadg (approved):
🙌

.github/workflows/lint-test.yml
samsiegart: I think so, according to https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07. Maybe we shouldn't?

samsiegart: I'll just merge it for now and we can take it out later if desired

Issue #16, opened 2022-12-01
As Vaults user managing my vault, I need to be able to borrow additional IST or repay existing IST debt from said vault (‘Adjust Debt’)


otoole-brendan:
## What is the Problem Being Solved?

Being able to adjust vault debt by repaying or borrowing more IST is a core function of vault design

## Description of the Design

Use Figma design for reference.

Should include below user interactions:

- User should be able to select ‘Borrow’ or ‘Repay’ from Action dropdown.

- User should be able to see their available IST balance to repay (or borrow more).

- User should be able to manually enter and/or max fill

## Security Considerations
Unknown
## Test Plan
Smoke Test after implementation

otoole-brendan:
GH spazzed out and created a duplicate of this issue so I close. Please ignore.

PR #63, opened 2023-02-10
feat: max input buttons for vault inputs

samsiegart:
refs #5 
refs #14 
refs #15 
refs #16

### DIFF
--- a/src/components/AdjustVaultForm.tsx
+++ b/src/components/AdjustVaultForm.tsx
@@ -7,6 +7,7 @@ import {
   DebtAction,
   debtActionAtom,
   debtDeltaValueAtom,
+  vaultAfterAdjustmentAtom,
   vaultToAdjustAtom,
 } from 'store/adjustVault';
 import { displayFunctionsAtom, pursesAtom } from 'store/app';
@@ -16,6 +17,13 @@ import StyledDropdown from './StyledDropdown';
 import { PursesJSONState } from '@agoric/wallet-backend';
 import CloseVaultDialog from './CloseVaultDialog';
 import { useState } from 'react';
+import { AmountMath } from '@agoric/ertp';
+import { makeRatioFromAmounts } from '@agoric/zoe/src/contractSupport';
+import {
+  addRatios,
+  floorDivideBy,
+  floorMultiplyBy,
+} from '@agoric/zoe/src/contractSupport/ratio';
 
 const AdjustVaultForm = () => {
   const displayFunctions = useAtomValue(displayFunctionsAtom);
@@ -26,6 +34,7 @@ const AdjustVaultForm = () => {
   const { displayBrandPetname } = displayFunctions;
 
   const vaultToAdjust = useAtomValue(vaultToAdjustAtom);
+  const vaultAfterAdjustment = useAtomValue(vaultAfterAdjustmentAtom);
 
   const [debtAction, setDebtAction] = useAtom(debtActionAtom);
   const [collateralAction, setCollateralAction] = useAtom(collateralActionAtom);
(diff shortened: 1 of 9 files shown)


src/store/createVault.ts
```diff
@@ -239,7 +243,18 @@ export const inputErrorsAtom = atom<VaultCreationErrors>(get => {
       selectedMetrics.totalDebt,
     ).value;
 
-    if (istAvailable < valueToReceive) {
```
samsiegart, line 242: This is incorrect because `valueToReceive` is how much IST you'll get, not how much debt you'll add. After multiplying it by loan fee the resulting debt will be higher.

src/components/AdjustVaultForm.tsx
```diff
@@ -49,6 +58,67 @@ const AdjustVaultForm = () => {
 
   const isActive = vaultToAdjust?.vaultState === 'active';
 
+  const onMaxCollateralClicked = () => {
+    assert(collateralAction === CollateralAction.Deposit);
```
turadg, line 62: helpful assertion. makes clear that when`actionLabel` is undefined the user can't click.

src/components/AdjustVaultForm.tsx
```diff
@@ -49,6 +58,67 @@ const AdjustVaultForm = () => {
...
+
+    if (!collateralPurse) {
+      /* no-op */
+      return;
+    }
+
+    setCollateralDeltaValue(collateralPurse.currentAmount.value);
```
turadg, line 69: I always read "delta" as a value that can be negative, like movement along a line. "change" sounds more like absolute value of the difference.

also it makes me nervous to be passing around the bigints instead of the typed amounts.

src/components/AdjustVaultForm.tsx
```diff
@@ -49,6 +58,67 @@ const AdjustVaultForm = () => {
...
+  };
+
+  const onMaxDebtClicked = () => {
```
turadg, line 72: most of this function belongs in `vaultMath` with unit tests

src/components/AdjustVaultSummary.tsx
```diff
@@ -80,7 +80,16 @@ const AdjustVaultSummary = () => {
   } = vaultToAdjust;
 
   const isActive = vaultState === 'active';
-  const offerButtonLabel = isActive ? 'Make Offer' : vaultState;
+
+  const offerSigner = useAtomValue(offerSignerAtom);
+
+  const offerButtonLabel = useMemo(() => {
+    if (!offerSigner?.isDappApproved) {
+      return 'Enable Dapp in Wallet';
```
turadg, line 88: is this what clicking the button will do? if not the button should be disabled and the user instruction go somewhere else.

src/components/AmountInput.tsx
```diff
@@ -1,32 +1,43 @@
...
+  {
+    value,
+    brand,
+    label,
+    error,
+    suffix,
+    actionLabel,
+    onChange = () => {
+      /* noop */
+    },
```
turadg, line 33: consider defining `noop` or importing it from Lodash.
```suggestion
    onChange = noop,
```

src/components/BaseDialog.tsx
```diff
@@ -21,6 +23,7 @@ const BaseDialog = ({
   secondaryActionLabel,
   onPrimaryAction,
   onSecondaryAction,
+  primaryActionDisabled = false,
```
turadg, line 26: given how tailored this component is to the actions use case, consider renaming it `ActionsDialog`.

src/components/ConfigureNewVault.tsx
```diff
@@ -66,6 +73,61 @@ const ConfigureNewVault = () => {
       ? `${displayBrandPetname(collateralBrand)} to lock up *`
       : 'To lock up *';
 
+  const purse = usePurseForBrand(collateralBrand);
+
+  const onMaxClicked = () => {
```
turadg, line 78: please extract to vaultMath with tests

src/store/createVault.ts
turadg: Thanks for the note. The commit shouldn't be a "fix" commit because it's fixing up other code in the same PR. Consider `fixup!` and using rebase to squash before merge. Or if you want to squash this whole PR nevermind this time.

The error points to the value of having math and numerical validation be separated from the UI components and have unit tests. Particularly around boundaries and thresholds.

src/components/ConfigureNewVault.tsx
samsiegart: Done

src/components/AdjustVaultSummary.tsx
samsiegart: No it's more of a help tip and the button should be disabled in this case. Just added an assert to make that clear.

src/components/AmountInput.tsx
samsiegart: Defined above.

src/components/BaseDialog.tsx
samsiegart: Done

src/components/AdjustVaultSummary.tsx
```diff
@@ -110,22 +102,45 @@ const AdjustVaultSummary = () => {
...
 
+  const isButtonDisabled = !canMakeOffer;
+
+  const offerButtonLabel = useMemo(() => {
+    if (!offerSigner?.isDappApproved) {
+      assert(
+        isButtonDisabled,
+        'Button should be disabled when dapp is not enabled in wallet',
+      );
+      return 'Enable Dapp in Wallet';
```
turadg, line 116: the disabling assertions help. it's still awkward to me that the button has an instruction. not a blocker.

src/components/AmountInput.tsx
```diff
@@ -8,6 +8,10 @@ import type { Brand } from '@agoric/ertp/src/types';
 import type { Ref } from 'react';
 import clsx from 'clsx';
 
+const noop = () => {
+  /* no-op */
```
turadg, line 12: 🤣

src/components/CollateralChoice.tsx
```diff
@@ -109,7 +109,7 @@ const CollateralChoice = ({ id, displayFunctions }: CollateralChoiceParams) => {
     displayBrandIcon,
   } = displayFunctions;
 
-  const istAvailable = AmountMath.subtract(params.debtLimit, metrics.totalDebt);
+  const mintedAvailable = istAvailable(params.debtLimit, metrics.totalDebt);
```
turadg, line 112: good call using role-based naming

src/service/vaults.ts
```diff
@@ -213,8 +213,17 @@ export const watchVaultFactory = (netconfigUrl: string) => {
           liquidationPadding,
         );
 
+      const debtLimit = current.DebtLimit.value;
+
+      // The real debt limit can actually never be reached, only 1 less than
```
turadg, line 218: Hopefully we can kill this code after https://github.com/Agoric/agoric-sdk/issues/6970

src/utils/vaultMath.ts
```diff
@@ -74,13 +74,13 @@ export const netValue = (lockedValue: Amount<'nat'>, debt: Amount<'nat'>) =>
     ? [AmountMath.subtract(lockedValue, debt), false]
     : [AmountMath.subtract(debt, lockedValue), true];
 
-export const debtAfterDelta = (
+export const debtAfterChange = (
   debtAction: DebtAction,
   loanFee: Ratio,
   totalDebt: Amount<'nat'>,
-  debtDeltaValue: bigint | null,
+  debtChange: Amount<'nat'> | null,
```
turadg, line 81: 👍

src/utils/vaultMath.ts
```diff
@@ -92,36 +92,118 @@ export const debtAfterDelta = (
...
+    desiredCollateralization,
+    istAvailableBeforeLoanFee.value,
+    desiredCollateralization,
+    loanFee,
+  );
+
+  return AmountMath.min(
+    AmountMath.make(collateralPurseBalance.brand, collateralForAvailableIst),
+    collateralPurseBalance,
+  ).value;
```
turadg, line 165: consider returning the Amount, so at the callsite it's clear that it cannot be neg.

ditto for other amount returns

turadg (approved):
Thanks for the test coverage

PR #67, opened 2023-02-11
feat: network config env var to replace dropdown

### LINKED ISSUE #36: remove mainnet from the repo
Refs: https://github.com/Agoric/agoric-sdk/issues/4350

The builds point to mainnet but we can't host, or deploy anything in CI, which points to mainnet (agoric-3).

## Description of the Design

One option is to introduce an environment variable, or alternate build flag to enable a "production" mode, which includes only the production config, so that default building of the project will only have testnet configurations

Another is to make this repo completely unaware of production uses. Make the endpoint selection be defined by a config file and when this code is purposed for a new network, anyone can modify the config file and build. One kink is that the config file is in the repo so they might trip up and commit their changes. To prevent that we could have config.ts read an optional JSON file for overrides. The override file would be listed in .gitignore so it doesn't ever get into the repo.

## Test Plan

No references to mainnet in the repo.

samsiegart:
closes #36

### DIFF
--- a/src/store/app.ts
+++ b/src/store/app.ts
@@ -114,10 +114,9 @@ export const leaderAtom = atom(
     })),
 );
 
-export const networkConfigAtom = atomWithStorage(
-  'agoric-network-config',
-  networkConfigs.mainnet,
-);
+export const networkConfigAtom = import.meta.env.VITE_NETWORK_CONFIG_URL
+  ? atom({ url: import.meta.env.VITE_NETWORK_CONFIG_URL, label: undefined })
+  : atomWithStorage('agoric-network-config', networkConfigs.localhost);
 
 const usp = new URLSearchParams(window.location.search);
 const wallet = usp.get('wallet');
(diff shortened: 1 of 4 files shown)


src/store/app.ts
```diff
@@ -114,10 +114,9 @@ export const leaderAtom = atom(
     })),
 );
 
-export const networkConfigAtom = atomWithStorage(
-  'agoric-network-config',
-  networkConfigs.mainnet,
-);
+export const networkConfigAtom = import.meta.env.VITE_NETWORK_CONFIG_URL
```
turadg, line 117: Needs documentation.

Why does it have VITE?

The name should convey that it's optional. Eg "FORCE_" to indicate that it overrides the UI config

samsiegart: I guess this PR also doubles as a design proposal, so apologies that it seems out of the blue.

Updated the readme, does that make it clear enough?

Re: VITE prefix, see  https://vitejs.dev/guide/env-and-mode.html#env-files -- "To prevent accidentally leaking env variables to the client, only variables prefixed with VITE_ are exposed to your Vite-processed code"

turadg (approved):

PR #69, opened 2023-02-13
fix: use max debt limit

samsiegart:
- Use full debt limit now that we have: https://github.com/Agoric/agoric-sdk/pull/6984
- Fix rounding error for "max" button in vault creation and added test.

### DIFF
--- a/src/utils/vaultMath.ts
+++ b/src/utils/vaultMath.ts
@@ -49,8 +49,6 @@ export const computeToLock = (
   toReceive: NatValue,
   defaultCollateralization: Ratio,
   loanFee: Ratio,
-  // Round up by default to preserve offer safety. But, if using this to set
-  // max input, round down to avoid hitting debt limit.
   remainderHandling: 'floor' | 'ceil' = 'floor',
 ): NatValue => {
   const multiply =
--- a/src/utils/vaultMath.ts
+++ b/src/utils/vaultMath.ts
@@ -49,7 +49,13 @@ export const computeToLock = (
   toReceive: NatValue,
   defaultCollateralization: Ratio,
   loanFee: Ratio,
+  // Round up by default to preserve offer safety. But, if using this to set
+  // max input, round down to avoid hitting debt limit.
+  roundDown: boolean = false,
 ): NatValue => {
+  const multiply = roundDown ? floorMultiplyBy : ceilMultiplyBy;
+  const divide = roundDown ? floorDivideBy : ceilDivideBy;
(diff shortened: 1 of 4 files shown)


test/utils/vaultMath.test.ts
```diff
@@ -232,30 +232,30 @@ describe('maxCollateralForNewVault', () => {
   it('should prevent the user from exceeding the mint limit', () => {
```
samsiegart, line 232: Confirmed this test now fails without change. We also have https://github.com/Agoric/dapp-inter/blob/6aa6db5ede8a9586d46fcd295dc50c78d1372ea6/test/utils/vaultMath.test.ts#L19 to test the round-up case.

src/utils/vaultMath.ts
```diff
@@ -49,7 +49,13 @@ export const computeToLock = (
   toReceive: NatValue,
   defaultCollateralization: Ratio,
   loanFee: Ratio,
+  // Round up by default to preserve offer safety. But, if using this to set
+  // max input, round down to avoid hitting debt limit.
+  roundDown: boolean = false,
```
turadg, line 54: technically it's not rounding. Let's be more explicit about ceil vs floor. https://github.com/Agoric/agoric-sdk/blob/bcf33feb2ff38d08714c77ef8e644409a4d894fa/packages/zoe/src/contractSupport/safeMath.js#L21-L27

Floor is the default way JS does bigint division (truncating non-integer) so it should be the default. 

```suggestion
  remainderHandling: 'floor' | 'ceil' = 'floor',
```

src/utils/vaultMath.ts
```diff
@@ -159,6 +165,7 @@ export const maxCollateralForNewVault = (
     istAvailableBeforeLoanFee.value,
     desiredCollateralization,
     loanFee,
+    true,
```
turadg, line 168: by using the string literal as the option it's readable here what it's doing.
```suggestion
    'ceil',
```

test/utils/vaultMath.test.ts
turadg: thanks for checking and noting

src/utils/vaultMath.ts
samsiegart: Done

src/utils/vaultMath.ts
samsiegart: Done

turadg (approved):

Issue #68, opened 2023-02-12
use `makeVaultInvitation` from vault manager instead of director


turadg:
## What is the Problem Being Solved?

The method is being removed from the director: https://github.com/Agoric/agoric-sdk/pull/6983

This will be needed for https://github.com/Agoric/agoric-sdk/issues/6518 also, though that's not scheduled.

## Description of the Design

This will require keeping track of the manager objects and targeting them.

For targeting, instead of this `contract` spec, https://github.com/Agoric/dapp-inter/blob/48fe77a6167d2182317e7097079c3b75fd10b966/src/service/vaults.ts#L361-L364
it wil have to be an `object` spec, blocking on https://github.com/Agoric/agoric-sdk/issues/6850

samsiegart:
>Look up the vault/collateral manager instance

Is this published in vstorage anywhere currently?

PR #70, opened 2023-02-14
feat: use new invitation spec

samsiegart:
fixes: https://github.com/Agoric/dapp-inter/issues/68

Tested manually opening, adjusting, and closing a vault.

Blocked: requires https://github.com/Agoric/wallet-app/pull/59 and https://github.com/Agoric/agoric-sdk/pull/6996

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "@agoric/casting": "^0.3.3-dev-1adb7a2.0",
     "@agoric/ertp": "^0.15.4-dev-1adb7a2.0",
     "@agoric/inter-protocol": "^0.13.2-dev-95a7fa5.0",
-    "@agoric/smart-wallet": "^0.4.3-dev-1adb7a2.0",
+    "@agoric/smart-wallet": "^0.4.3-dev-8cbf697.0",
     "@agoric/ui-components": "^0.3.4-dev-1adb7a2.0",
     "@agoric/wallet": "^0.17.1-dev-1adb7a2.0",
     "@agoric/web-components": "^0.5.1-dev-1adb7a2.0",
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "preview": "vite preview",
-    "lint": "eslint src",
-    "format": "prettier --write . && yarn lint --fix",
+    "lint": "eslint . && prettier --check .",
+    "lint:fix": "eslint . --fix && prettier --write .",
     "test": "vitest",
     "coverage": "vitest run --coverage",
     "postinstall": "patch-package"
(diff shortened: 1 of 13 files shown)


src/service/vaults.ts
```diff
@@ -354,13 +332,19 @@ export const makeOpenVaultOffer = async (
...
-  const INVITATION_METHOD = 'makeVaultInvitation';
-
-  const { vaultFactoryInstanceHandle } = useVaultStore.getState();
   const { importContext, offerSigner } = appStore.getState();
-  const serializedInstance = importContext.fromBoard.serialize(
-    vaultFactoryInstanceHandle,
-  ) as CapData<'Instance'>;
+
+  const invitationSpec = importContext.fromBoard.serialize(
+    harden({
```
turadg, line 338: consider annotating the type here to statically validate the shape

src/service/vaults.ts
```diff
@@ -418,14 +401,16 @@ export const makeAdjustVaultOffer = async ({
...
 }: AdjustParams) => {
-  const invitationSpec = {
-    source: 'continuing',
-    previousOffer: vaultOfferId,
-    invitationMakerName: 'AdjustBalances',
-  };
-
   const { importContext, offerSigner } = appStore.getState();
 
+  const invitationSpec = importContext.fromBoard.serialize(
```
turadg, line 406: why does this need to be serialized now?

turadg (commented):
Love to see all the red!

I had a question before approval.

Also there's a lint warning on vaults.ts for an unused type. I think we should make that be an error but not a blocker.

src/service/vaults.ts
samsiegart: The wallet app now expects the invitation spec to be capdata. This supports the `agoricContract` proposal which has a `Brand` as an arg. It was easier this way than making the wallet-app check if it's capdata or not.

turadg: ok, makes sense that what goes between the apps is marshaled.

turadg (approved):

package.json
```diff
@@ -8,7 +8,8 @@
     "build": "tsc && vite build",
     "preview": "vite preview",
     "lint": "eslint src",
-    "format": "prettier --write . && yarn lint --fix",
+    "prettier": "prettier --check src",
```
turadg, line 11: We don't have a separate script for this in other packages. It's a sort of lint so we have "lint" do it. Consider aligning with other repos.

package.json
```diff
@@ -8,7 +8,8 @@
...
+    "format": "prettier --write src && yarn lint --fix",
```
turadg, line 12: Format should only affect white space and eslint fixers do more than that. They can break code. Please remove eslint from format.

Also, why only src? A bunch of non-src files were reformatted in this PR. Please omit those changes or make this continue to reformat them.

turadg (approved):
Some requests that I trust you'll do before merge. Approving now given dependencies so you don't need another review to land.

package.json
samsiegart: Done

package.json
samsiegart: Done

PR #71, opened 2023-02-14
feat: allow queryparam netselect

arirubinstein:
Allows a `network=ollinet` or default for network selection

### DIFF
--- a/src/components/NetworkDropdown.tsx
+++ b/src/components/NetworkDropdown.tsx
@@ -32,9 +32,11 @@ const Item = ({
 
 const NetworkDropdown = () => {
   const [networkConfig, setNetworkConfig] = useAtom(networkConfigAtom);
-  const specifiedNetworkName = new URLSearchParams(window.location.search).get('network');
+  const specifiedNetworkName = new URLSearchParams(window.location.search).get(
+    'network',
+  );
   if (specifiedNetworkName !== null) {
-    if (specifiedNetworkName in networkConfigs){
+    if (specifiedNetworkName in networkConfigs) {
       setNetworkConfig((networkConfigs as never)[specifiedNetworkName]);
       const prevSearch = new URLSearchParams(window.location.search);
       prevSearch.delete('network');
@@ -43,7 +45,7 @@ const NetworkDropdown = () => {
       console.error('invalid network requested', specifiedNetworkName);
     }
   }
-  
+
   const items = Object.values(networkConfigs).map(config => (
     <Item
       key={config.url}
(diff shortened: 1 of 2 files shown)


src/components/NetworkDropdown.tsx
```diff
@@ -32,7 +32,14 @@ const Item = ({
 
 const NetworkDropdown = () => {
   const [networkConfig, setNetworkConfig] = useAtom(networkConfigAtom);
-
+  const specifiedNetworkName = new URLSearchParams(window.location.search).get('network');
+  if (specifiedNetworkName !== null && specifiedNetworkName in networkConfigs) {
```
turadg, line 36: consider notifying by `console.error` if someone passing a `network` and it isn't defined in networkConfigs

src/components/NetworkDropdown.tsx
```diff
@@ -32,7 +32,14 @@ const Item = ({
...
+    setNetworkConfig((networkConfigs as never)[specifiedNetworkName]);
```
turadg, line 37: Does this persist in localstorage? If it's in-memory it'll be wiped out when `window.location.replace` does a navigation.

Either way, better to use [pushState](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState) that a navigation.

src/components/NetworkDropdown.tsx
samsiegart, line 36: We should also only do this if the url param is different than the currently selected networkConfig.

samsiegart (changes requested):

src/components/NetworkDropdown.tsx
```diff
@@ -32,7 +32,18 @@ const Item = ({
 
 const NetworkDropdown = () => {
   const [networkConfig, setNetworkConfig] = useAtom(networkConfigAtom);
-
+  const specifiedNetworkName = new URLSearchParams(window.location.search).get('network');
+  if (specifiedNetworkName !== null) {
+    if (specifiedNetworkName in networkConfigs){
```
samsiegart, line 37: If you're using vscode you can install eslint and prettier plugins to auto-lint and format. Otherwise you'll have to run `yarn format`

samsiegart: I guess our CI checks lint but not format, I'll have to add that

src/components/NetworkDropdown.tsx
arirubinstein: this gets a bit funny if you then want to change networks, as the url param will force you back to the original one.

samsiegart: I guess we always want to clear the param. We could use https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState instead of reloading the page in that case but probably not worth the complexity

samsiegart (approved):

PR #80, opened 2023-03-02
Fix Some Issues from VFT

### LINKED ISSUE #73: Make Collateralization Ratio field not editable
Current Vaults UI design lets users edit all 3 fields under 'Configure' when creating a vault. This UX is excessive and could potentially frustrate users who are likely to not adjust based on CR alone but likely just on a) the ATOM they want to lock up or b) the target IST they want to receive.
Making 'Collateralization Percent' field not editable simplifies the UX. 

Note: The field should still update dynamically depending on what user enters for other two fields.

## Description of the Design

See discussion here for latest: https://agoricopco.slack.com/archives/C03FUCJV2J2/p1676491047178079

High-level:
-Entering ATOM wouldn't update IST
-Max button on ATOM would input your full purse balance
-Max button on IST would input 1/17.5 of you current ATOM input value up to the remaining limit, whichever is less

## Security Considerations
TBD
## Test Plan

Will be smoke-tested

### LINKED ISSUE #74: Display Current Collateral Price during Vault Creation
When creating a vault, users cannot see the collateral price being used to determine collateralization percent and therefore IST debt. This will be the lower of either current market price per Oracle or the next liquidation price ([per ticket](https://github.com/Agoric/agoric-sdk/issues/6968)).

## Description of the Design
Display relevant collateral price once the user selects collateral type.

Show current price to user - suggest it should appear under and left of 'Cancel' button - in line with 'ATOM' (or relevant brand name) here https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=2322%3A11419&t=6nO7OvoKbHCwFhg4-0 to match location for displaying it during 'Adjust Vault'.

Use same style/design as 'Adjust Vaults'.
## Security Considerations
TBD
## Test Plan
Will be smoke tested

### LINKED ISSUE #76: Update Adjust Vault view with latest changes
Following implementation this is a list of minor UI changes following testing that should be applied prior to launch

## Description of the Design

- [ ] Outstanding Debt in vault summary bar currently shows as 'US$', this should be IST as debt denomination is always in IST in the system, unless we specify US$ value of said debt

## Security Considerations
N/A
## Test Plan
Will be smoke-tested

### LINKED ISSUE #77: Open existing Wallet UI browser tab when I click 'Go To Wallet' dialog after creating a vault
After I hit 'Create Vault' button in Vaults UI, a dialog appears to let me know my vault creation request has been submitted and that I need to approve in my smart wallet. 
When I click 'GO TO WALLET' it should open the existing Wallet UI browser tab that I already have open, not open a new one (thus resulting in me having two Wallet UI tabs)

## Description of the Design

As described above. Detect if user has Wallet UI open in another browser tab and navigate there rather than opening a new tab.

### LINKED ISSUE #81: Remove 'Swaps' reference on 'Dapp is in Read-only mode' toast
After you've instantiated a wallet and returned to the Vaults UI - a user will see a PSM-era toast letting them know they are in read only mode and that they need to approve dapp in wallet UI to perform 'swaps'.

'Swaps' are a PSM term and aren't relevant for Vaults functionality so need to be removed/updated.

## Description of the Design

See current toast:

This needs to be updated to ' ...enable the connection at X to continue interacting with the Vaults UI'
## Security Considerations
N/A
## Test Plan
Will be smoke tested

samsiegart:
fixes #77 
fixes #76 
fixes #74 
fixes #73
fixes #81 

image

### DIFF
--- a/src/components/ConfigureNewVault.tsx
+++ b/src/components/ConfigureNewVault.tsx
@@ -23,7 +23,7 @@ const ConfigureNewVault = () => {
     prices: vaults.prices,
   }));
 
-  const { displayPercent, displayBrandPetname } =
+  const { displayPercent, displayBrandPetname, displayPrice } =
     useAtomValue(displayFunctionsAtom) ?? {};
 
   const [valueToLock, setValueToLock] = useAtom(valueToLockAtom);
@@ -67,6 +67,14 @@ const ConfigureNewVault = () => {
       ? `${displayBrandPetname(collateralBrand)} to lock up *`
       : 'To lock up *';
 
+  const collateralPriceForDisplay =
+    displayBrandPetname &&
+    displayPrice &&
+    hasPriceFeed &&
+    `1 ${displayBrandPetname(collateralBrand)} = ${displayPrice(
+      prices.get(collateralBrand),
+    )}`;
+
   const purse = usePurseForBrand(collateralBrand);
 
   const onMaxClicked = () => {
@@ -102,8 +110,11 @@ const ConfigureNewVault = () => {
       <p className="font-serif text-[#666980] leading-[26px]">
         Choose your vault parameters.
       </p>
-      <div className="mt-4 mb-4 text-sm font-serif text-[#666980]">
-        <span className="font-bold">{purseBalance}</span> Available
+      <div className="mt-4 mb-4 flex flex-wrap gap-x-[156px] gap-y-4 text-sm font-serif text-[#666980]">
+        <div>
+          <span className="font-bold">{purseBalance}</span> Available
+        </div>
+        <div>{collateralPriceForDisplay}</div>
       </div>
       <div className="flex gap-x-20 gap-y-6 flex-wrap">
         <AmountInput
--- a/src/components/ConfigureNewVault.tsx
+++ b/src/components/ConfigureNewVault.tsx
@@ -1,3 +1,4 @@
+import { AmountMath } from '@agoric/ertp';
 import AmountInput from 'components/AmountInput';
 import RatioPercentInput from 'components/RatioPercentInput';
 import { useAtom, useAtomValue } from 'jotai';
@@ -11,7 +12,7 @@ import {
 } from 'store/createVault';
 import { useVaultStore } from 'store/vaults';
 import { usePurseBalanceDisplay, usePurseForBrand } from 'utils/hooks';
-import { maxCollateralForNewVault } from 'utils/vaultMath';
+import { maxIstToBorrowFromVault } from 'utils/vaultMath';
 
 const ConfigureNewVault = () => {
   const { collateralizationRatioError, toLockError, toReceiveError } =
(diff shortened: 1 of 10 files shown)


src/components/ConfigureNewVault.tsx
```diff
@@ -114,12 +125,14 @@ const ConfigureNewVault = () => {
...
+          onAction={onMaxCollateralClicked}
         />
         <RatioPercentInput
-          onChange={setCollateralizationRatio}
+          onChange={() => {
+            /* always disabled */
+          }}
           value={collateralizationRatio}
-          disabled={!isInputReady}
+          disabled={true}
```
turadg, line 135: I see this implements https://github.com/Agoric/dapp-inter/issues/73 as spec-ed, but I have doubts about the UX of an uneditable input field. Maybe we need a UI pattern for "dynamic value that you don't control". I see three kinds of numbers in the UI,
- user sets
- governance sets
- market sets
…and of course composites of those.

@otoole-brendan WDYT of having some visual queue to categorize them as such? If the categories were colors, the composite could be blends.

src/service/wallet.tsx
```diff
@@ -1,3 +1,4 @@
+/** This service is a .tsx so it can present contentful toasts. */
```
turadg, line 1: helpful comment. though I worry we'll put more and more JSX in the file. solution below.

src/service/wallet.tsx
```diff
@@ -103,7 +104,7 @@ export const makeWalletService = () => {
           showToast('Network not found.');
           break;
         case Errors.noSmartWallet:
-          showToast(SmartWalletNotFoundToast);
+          showToast(<SmartWalletNotFoundToast />);
```
turadg, line 107: it's `showToast`'s job to render. It should be able to take a component without JSX. i.e.,
```
  const showToast = (component: ReactComponent, options?: ToastOptions) => {
    const content: ToastContent = React.createElement(component, null);
    toastId = toast.error(content, options);
  };
```
Then this file can stay `.ts` and this line needn't change.

turadg (approved):
Approving conditional on reverting the `.tsx` change. If you disagree let's chat.

src/components/ConfigureNewVault.tsx
otoole-brendan: @turadg not sure I follow what you mean by 'visual queue to categorize' - are you saying that anything a user sets is denoted by a global color, anything that is a governance parameter set (and not editable) is a different global color and market set i.e. ATOM price is a different global color again?

turadg: By way of illustration, yes. Color is an example of a visual cue (misspelled earlier), which has the nice blending affordance, but I'm not suggesting that necessarily. Just that there be some consistent cue of what controls the number. Another could be,
- user numbers are in an input box
- governance numbers have a highlighted background
- market numbers have a little dot next to them to show how long ago they were read

src/service/wallet.tsx
samsiegart: Very nice suggestion, done. However, I'm leaving the interface for `showToast` unchanged because `ToastContent` can also be a string, not always a component. I think of this as a small wrapper on `toast.error` just to persist the toast id.

src/service/wallet.ts
```diff
@@ -1,4 +1,4 @@
-/** This service is a .tsx so it can present contentful toasts. */
```
turadg: consider squashing this commit into the one that made this line so it doesn't exist in master history.

turadg (approved):

src/service/wallet.ts
samsiegart: 👍Done

PR #89, opened 2023-03-16
fix: convert entries from wallet

samsiegart:
Vaults weren't loading because we changed the wallet to publish an array of entries rather than a record for its public subscribers [#Agoric/agoric-sdk/7139](https://github.com/Agoric/agoric-sdk/pull/7139). This updates the UI to view parse them correctly.


dckc (approved):
works for me!

image

PR #91, opened 2023-03-24
fix: show error when withdrawing more than locked

samsiegart:
image

### DIFF
--- a/src/store/adjustVault.ts
+++ b/src/store/adjustVault.ts
@@ -207,6 +207,11 @@ export const vaultAfterAdjustmentAtom = atom<VaultAfterAdjustment | null>(
   },
 );
 
+/**
+ * Only returns a maximum of one error each for collateral and debt, in no
+ * particular order of precedence. The important part is the user sees an error
+ * if one exists.
+ */
 export const adjustVaultErrorsAtom = atom(get => {
   let debtError;
   let collateralError;
@@ -257,6 +262,7 @@ export const adjustVaultErrorsAtom = atom(get => {
   const collateralPurseAmount = collateralPurse?.currentAmount as
     | Amount<'nat'>
     | undefined;
+
   if (
     collateralAction === CollateralAction.Deposit &&
     collateralInputAmount?.value &&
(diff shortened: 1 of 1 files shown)


src/store/adjustVault.ts
```diff
@@ -266,6 +266,13 @@ export const adjustVaultErrorsAtom = atom(get => {
     collateralError = 'Insufficient funds.';
   }
 
+  if (
+    collateralAction === CollateralAction.Withdraw &&
+    AmountMath.isEmpty(newLocked)
+  ) {
+    collateralError = 'Exceeds available collateral.';
```
turadg, line 273: could there be this error and others concurrently?

src/store/adjustVault.ts
```diff
@@ -266,6 +266,13 @@ export const adjustVaultErrorsAtom = atom(get => {
     collateralError = 'Insufficient funds.';
   }
 
+  if (
+    collateralAction === CollateralAction.Withdraw &&
+    AmountMath.isEmpty(newLocked)
```
turadg, line 271: why does empty newLocked imply excess? wouldn't it also be zero if you withdraw exactly all of it?

src/store/adjustVault.ts
samsiegart: Multiple error conditions could be true simultaneously, but `collateralError` only holds one. I don't think it's worth the overhead and noisiness of showing multiple errors at once.

src/store/adjustVault.ts
samsiegart: The contract won't let you adjust a vault to have 0 collateral, so effectively the "available collateral" to withdraw is 1 less than the amount locked. This was succinct, but we could probably say something like "This would empty the vault completely. Close the vault [with a link to the close action] instead?"

turadg: Makes sense. Consider documenting with comment

src/store/adjustVault.ts
turadg: Agree. It's not clear to me if the error message precedence here is intentional or not. Would be important to know when refactoring, especially since there aren't tests to imply a spec.

turadg (approved):

src/store/adjustVault.ts
samsiegart: Done

src/store/adjustVault.ts
samsiegart: Yea makes sense, just added some documentation to make it less unclear.

samsiegart:
TY4R!

PR #104, opened 2023-04-27
feat: legal disclaimer

### LINKED ISSUE #54: Add Legal Disclaimer to Vaults UI
Similar to PSM and Wallet UI, we need to serve users with a legal disclaimer upon accessing the UI for the first time. User must accept to be able to continue
## Description of the Design

Same design as PSM: https://github.com/Agoric/agoric-sdk/issues/6491 . Copy TBD once Legal confirms.

## Security Considerations
N/A
## Test Plan
To be tested during smoke testing

samsiegart:
fixes #54 

image

### DIFF
--- a/src/components/ActionsDialog.tsx
+++ b/src/components/ActionsDialog.tsx
@@ -2,16 +2,19 @@ import { Dialog, Transition } from '@headlessui/react';
 import clsx from 'clsx';
 import { Fragment, ReactElement } from 'react';
 
+type DialogAction = {
+  label: string;
+  action: () => void;
+};
+
 type Props = {
   isOpen: boolean;
   onClose: () => void;
-  onPrimaryAction: () => void;
   title: string;
   body: ReactElement;
-  primaryActionLabel: string;
+  primaryAction: DialogAction;
+  secondaryAction?: DialogAction;
   primaryActionDisabled?: boolean;
-  secondaryActionLabel?: string;
-  onSecondaryAction?: () => void;
 };
 
 const ActionsDialog = ({
@@ -19,10 +22,8 @@ const ActionsDialog = ({
   onClose,
   title,
   body,
-  primaryActionLabel,
-  secondaryActionLabel,
-  onPrimaryAction,
-  onSecondaryAction,
+  primaryAction,
+  secondaryAction,
   primaryActionDisabled = false,
 }: Props) => {
   return (
@@ -61,12 +62,12 @@ const ActionsDialog = ({
                 <div className="h-[1px] mx-8 bg-[#D8D8D8]" />
                 <div className="py-6 px-8">
                   <div className="flex justify-end gap-6">
-                    {secondaryActionLabel && onSecondaryAction && (
+                    {secondaryAction && (
                       <button
                         className="text-btn-xs flex justify-center rounded  text-[#A3A5B9] border-[#A3A5B9] border-2 px-6 py-3 bg-gray-500 bg-opacity-0 hover:bg-opacity-10 active:bg-opacity-20"
-                        onClick={onSecondaryAction}
+                        onClick={secondaryAction.action}
                       >
-                        {secondaryActionLabel}
+                        {secondaryAction.label}
                       </button>
                     )}
                     <button
@@ -77,9 +78,9 @@ const ActionsDialog = ({
                           ? 'bg-gray-400 cursor-not-allowed'
                           : 'bg-interPurple hover:opacity-80 active:opacity-60',
                       )}
-                      onClick={onPrimaryAction}
+                      onClick={primaryAction.action}
                     >
-                      {primaryActionLabel}
+                      {primaryAction.label}
                     </button>
                   </div>
                 </div>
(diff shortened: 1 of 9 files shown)


src/components/ActionsDialog.tsx
```diff
@@ -61,12 +61,14 @@ const ActionsDialog = ({
                 <div className="h-[1px] mx-8 bg-[#D8D8D8]" />
                 <div className="py-6 px-8">
                   <div className="flex justify-end gap-6">
-                    <button
-                      className="text-btn-xs flex justify-center rounded  text-[#A3A5B9] border-[#A3A5B9] border-2 px-6 py-3 bg-gray-500 bg-opacity-0 hover:bg-opacity-10 active:bg-opacity-20"
-                      onClick={onSecondaryAction}
-                    >
-                      {secondaryActionLabel}
-                    </button>
+                    {secondaryActionLabel && onSecondaryAction && (
```
turadg, line 64: The type makes them independently optional but this is showing they go together. Consider making the type like `secondaryAction?: [label: string, onPress: () => void]`

src/components/TermsDialog.tsx
```diff
@@ -0,0 +1,70 @@
...
+          checked={isChecked}
+          onClick={() => setIsChecked(isChecked => !isChecked)}
+        />
+        <span>
+          By clicking here you are indicating that you have read and agree to
+          our{' '}
+          <a
+            className="text-blue-500 hover:text-blue-700 underline transition-colors"
+            target="inter_psm_disclaimer"
+            href="https://docs.inter.trade/disclaimer"
```
turadg, line 43: may change independently of behavior. should be a const. maybe in a config module

src/components/TermsDialog.tsx
```diff
@@ -0,0 +1,70 @@
...
+            Disclaimer
+          </a>
+          .
+        </span>
+      </label>
+    </div>
+  );
+
+  return (
+    <>
```
turadg, line 54: unnecessary nesting

src/service/wallet.ts
```diff
@@ -84,6 +89,13 @@ export const makeWalletService = () => {
 
     if (isWalletConnectionInProgress || chainConnection) return;
 
+    if (
+      latestTermsIndex !== localStorageStore.getState().latestTermsAgreedIndex
```
turadg, line 93: consider having something in the dialog (maybe in small gray print) that the terms must be agreed to in each browser. to stave off support requests about having to agree repeatedly. 

do we check anywhere that localStorage is available? we might need to error when it's disabled. people may auto-clear it but not much we can do about that

src/store/app.ts
```diff
@@ -134,3 +143,22 @@ export const walletUiHrefAtom = atom(get => {
 
   return bridgeUrl ? bridgeUrl.origin + '/wallet/' : '';
 });
+
+// Increment whenever the terms on https://docs.inter.trade/disclaimer change.
+export const latestTermsIndex = 0;
```
turadg, line 148: "Terms" has a meaning in our contract API. consider "Disclaimer" or "[ToS](https://en.wikipedia.org/wiki/Terms_of_service)"

yarn.lock
```diff
@@ -6643,9 +6643,9 @@ yocto-queue@^0.1.0:
   resolved "https://registry.yarnpkg.com/yocto-queue/-/yocto-queue-0.1.0.tgz#0294eb3dee05028d31ee1a5fa2c556a6aaf10a1b"
   integrity sha512-rVksvsnNCdJ/ohGc6xgPwyN8eheCxsiLM8mxuE/t/mOVqJewPuO1miLpTHQiRgTKCLexL4MeAFVagts7HmNZ2Q==
 
-zustand@^4.1.5:
-  version "4.1.5"
-  resolved "https://registry.yarnpkg.com/zustand/-/zustand-4.1.5.tgz#7402b511f5b23ccb0f9ba6d20ae01ec817e16eb6"
-  integrity sha512-PsdRT8Bvq22Yyh1tvpgdHNE7OAeFKqJXUxtJvj1Ixw2B9O2YZ1M34ImQ+xyZah4wZrR4lENMoDUutKPpyXCQ/Q==
+zustand@^4.3.7:
```
turadg, line 6646: curious, was this necessary or just tidying up?

for changelog hygiene please try to keep dep bumps in their own commits

turadg (approved):

yarn.lock
samsiegart: Neither, I was just trying to get it to work and it wasn't actually needed. Reverted this change for now.

src/service/wallet.ts
samsiegart: I'm not sure exactly how to word that, especially considering multiple ipfs hash domains, but I can follow up and see if that's more helpful or just more confusing.

Looks like if localstorage isn't available, the app just crashes because `atomWithStorage` throws. I'm reluctant to try and handle that because it's pretty out of the ordinary, and at least for chrome, you'll see a "site not working?" tip whenever it's off by default.

src/components/ActionsDialog.tsx
samsiegart: Done. I was just hesitant about un-flattening the params because it impacts equality checking, but for this it's not much impact.

PR #105, opened 2023-04-29
feat: support new managers vstorage path

turadg:
required for https://github.com/Agoric/agoric-sdk/pull/7150

Support both before and after schemes. We can drop support for the old scheme when it's known not to be served.


samsiegart (approved):
Thanks, works with my local chain so no regressions afaik.

PR #107, opened 2023-05-02
fix: support new vstorage paths

samsiegart:
Should be backwards compatible with previous paths


src/service/vaults.ts
```diff
@@ -88,7 +95,14 @@ type VaultSubscribers = {
  * Ex. `asset: 'published.vaultFactory.manager0'` -> `'manager0'`.
  */
 const getManagerIdFromSubscribers = (subscribers: VaultSubscribers) =>
-  subscribers.vault.split('.').find(node => node.includes('manager'));
+  subscribers.vault
+    .split('.')
+    .find(node => node.startsWith('manager') && !node.startsWith('managers'));
```
turadg, line 100: Given the example, `lastNode` would work but I take it this is used for longer paths like `published.vaultFactory.managers.manager0.metrics`. That would be good to include in the example.

Consider using a regex for further clarity. And refrain from "node" since that's the kind of the object for writing and this function is just string manipulation.
```suggestion
    .find(segment => segment.match(MANAGER_RE));
```
where with this in module scope:
```js
const MANAGER_RE = /manager\d+/;
```

Alternately (not recommended):
```suggestion
    .find(MANAGER_RE.test.bind(MANAGER_RE));
```

src/service/vaults.ts
```diff
@@ -88,7 +95,14 @@ type VaultSubscribers = {
...
+
+const lastNode = (path: string) => {
```
turadg, line 102: Similarly to the "node" comment above. We're not getting out a node object, just a segment of the `path` param.
```suggestion
const lastSegment = (path: string) => {
```

src/service/vaults.ts
```diff
@@ -249,11 +266,11 @@ export const watchVaultFactory = (netconfigUrl: string) => {
     }
   };
 
-  const watchManager = async (id: string) => {
-    watchGovernedParams(id);
-    watchMetrics(id);
+  const watchManager = async (path: string) => {
```
turadg, line 269: better for this to take the manager id as it had been, but I see the watchers for the two subpaths wouldn't know which scheme to read from. So I think on balance this is good. But then for robustness and clarity assert the `id` matches `MANAGER_RE`.

src/service/vaults.ts
```diff
@@ -290,19 +307,21 @@ export const watchVaultFactory = (netconfigUrl: string) => {
...
     let managerIds: string[];
+    let managerPrefix = 'published.vaultFactory';
     try {
       // old way (deprecated since https://github.com/Agoric/agoric-sdk/pull/7150)
-      managerIds = await fetchVstorageKeys(rpc, 'published.vaultFactory').then(
-        res =>
-          (res.children as string[]).filter(key => key.startsWith('manager')),
+      managerIds = await fetchVstorageKeys(rpc, managerPrefix).then(res =>
+        (res.children as string[]).filter(
+          key => key.startsWith('manager') && !key.startsWith('managers'),
```
turadg, line 315: MANAGER_RE

turadg (approved):
Thanks for making it fully backwards compatible.

src/service/vaults.ts
samsiegart: Thanks for the regex, I normally try to avoid it for readability but this one seems simple enough.

>I take it this is used for longer paths like published.vaultFactory.managers.manager0.metrics

I was careful not to call it on that, only on the `prefix`, but moot point now.

src/service/vaults.ts
samsiegart: Added an assert in the util fn

src/service/vaults.ts

src/service/vaults.ts

PR #110, opened 2023-05-04
Update README.md

samsiegart:
Hint to read ahead for `--interactive` flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,14 @@ User application for Agoric Inter Protocol--Vaults, BLD Boost, Liquidations, etc
 
    `agd --help`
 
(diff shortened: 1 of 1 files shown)


samsiegart:
> Better as a footnote[1](#user-content-fn-1-21fc0fd495ad78faee0dcff2baa229ca)
> 
> ## Footnotes
> 1. it's a Markdown feature https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/ [↩](#user-content-fnref-1-21fc0fd495ad78faee0dcff2baa229ca)

That's neat but I think the footnote isn't prominent enough so I'll just leave it as is

PR #111, opened 2023-05-05
fix: use new price timestamp format

samsiegart:
The app crashes now on `displayPriceTimestamp`, this fixes it.

### DIFF
--- a/src/utils/displayFunctions.ts
+++ b/src/utils/displayFunctions.ts
@@ -122,7 +122,7 @@ export const makeDisplayFunctions = (brandToInfo: Map<Brand, BrandInfo>) => {
     return new Intl.DateTimeFormat(navigator.language, {
       timeStyle: 'medium',
       dateStyle: 'short',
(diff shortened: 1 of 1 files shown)


src/utils/displayFunctions.ts
```diff
@@ -122,7 +122,7 @@ export const makeDisplayFunctions = (brandToInfo: Map<Brand, BrandInfo>) => {
     return new Intl.DateTimeFormat(navigator.language, {
       timeStyle: 'medium',
       dateStyle: 'short',
-    }).format(new Date(Number(price.timestamp) * 1000));
+    }).format(new Date(Number(price.timestamp.absValue) * 1000));
```
turadg, line 125: This could be backward compatible by using the `absValue` helper in agoric/time. (Named something like that)

If there are other time values this may be a good time(!) to wrap them with the accessor function (because it supports both formats)

turadg (approved):

PR #118, opened 2023-05-16
feat: account for locked price in creation and adjustment

### LINKED ISSUE #100: User cannot see next liquidation price when creating a vault

We need to train users to be cognizant of the next liquidation price when creating a vault so they can adequately collateralize at time of creation (give themselves enough buffer).

Showing this enables them to factor this in

## Description of the Design

Add 'Next Liquidation Price' (take Locked Price) to vault configure box - only show when price lock period = active

## Security Considerations
N/A
## Test Plan
To be smoke tested

### LINKED ISSUE #101: User can unknowingly create/adjust a vault that quickly becomes liquidated
UI work that compliments #6968 - see for more information that is adequately described there.

## Description of the Design

Vaults UI work for underlying contract work spec'd as https://github.com/Agoric/agoric-sdk/issues/6968 

(use lower price between current oracle price or next liquidation price when creating a vault or adjusting a vault (minting more IST) in lieu of MCR-LR calculation in avoid unwitting liquidation situations.

samsiegart:
fixes #100 
fixes #101 

## Description

Makes the "max" button use the lowest of next liquidation/current quote price when creating and adjusting vaults, and enforces collat. ratio based on lowest price when creating a vault.

Shows next liquidation price during price lock period in creation and adjustment views:

image

Capture2

## Testing

Tested manually throughout price lock periods on a local chain and added unit tests for new cases

### DIFF
--- a/src/utils/hooks.ts
+++ b/src/utils/hooks.ts
@@ -1,7 +1,8 @@
 import { useMemo } from 'react';
 import { useAtomValue } from 'jotai';
-import { displayFunctionsAtom, pursesAtom } from 'store/app';
+import { currentTimeAtom, displayFunctionsAtom, pursesAtom } from 'store/app';
 import type { Amount, Brand } from '@agoric/ertp/src/types';
+import { LiquidationSchedule } from 'store/vaults';
 
 export const usePurseBalanceDisplay = (brand?: Brand<'nat'> | null) => {
   const purses = useAtomValue(pursesAtom);
@@ -29,4 +30,27 @@ export const usePurseForBrand = (brand?: Brand<'nat'> | null) => {
   const purses = useAtomValue(pursesAtom);
 
   return purses?.find(p => p.brand === brand);
+};
+
+export const useAuctionTimer = (schedule: LiquidationSchedule | null) => {
+  const currentTime = useAtomValue(currentTimeAtom);
+
+  const minutesUntilNextAuction =
+    schedule?.nextStartTime &&
+    Math.max(
+      Math.floor((Number(schedule.nextStartTime.absValue) - currentTime) / 60),
+      0,
+    );
+
+  const secondsUntilNextAuction =
+    schedule?.nextStartTime &&
+    Math.max(
+      Math.floor((Number(schedule.nextStartTime.absValue) - currentTime) % 60),
+      0,
+    );
+
+  return minutesUntilNextAuction !== undefined &&
+    secondsUntilNextAuction !== undefined
+    ? `in ${minutesUntilNextAuction}m ${secondsUntilNextAuction}s`
+    : '';
 };
--- a/src/utils/vaultMath.ts
+++ b/src/utils/vaultMath.ts
@@ -33,7 +33,7 @@ export const isLiquidationPriceBelowGivenPrice = (
 };
 
 export const computeToLock = (
-  priceRate: Ratio,
+  quotePrice: Ratio,
   collateralizationRatio: Ratio,
   toReceive: NatValue,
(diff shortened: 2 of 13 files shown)


src/utils/hooks.ts
```diff
@@ -30,3 +31,26 @@ export const usePurseForBrand = (brand?: Brand<'nat'> | null) => {
...
+    schedule?.nextStartTime &&
+    Math.max(
+      Math.floor((Number(schedule.nextStartTime.absValue) - currentTime) % 60),
+      0,
+    );
+
+  return minutesUntilNextAuction !== undefined &&
+    secondsUntilNextAuction !== undefined
+    ? `in ${minutesUntilNextAuction}m ${secondsUntilNextAuction}s`
+    : '';
```
turadg, line 55: do you really mean to treat schedule in the past as at zero? I think it might be better to show the negative.

consider,
```suggestion
  const secondsLeft =
    schedule?.nextStartTime &&
    Math.floor(Number(schedule.nextStartTime.absValue) - currentTime);

  if (!secondsLeft) {
    return '';
  }

  return `in ${secondsLeft / 60}m ${secondsLeft % 60}s`;
```

src/utils/vaultMath.ts
```diff
@@ -119,6 +122,9 @@ export const lockedAfterChange = (
   return AmountMath.makeEmpty(locked.brand);
 };
 
+const lowestPrice = (priceA: Ratio, priceB?: Ratio) =>
+  priceB ? (ratioGTE(priceA, priceB) ? priceB : priceA) : priceA;
```
turadg, line 126: this would be handy in `ratio.js`

src/utils/vaultMath.ts
```diff
@@ -174,11 +184,12 @@ export const collateralizationRatio = (
   price: PriceDescription,
```
turadg, line 184: now that there are two price params, they should be distinguished. e.g. `marketPrice`

test/utils/vaultMath.test.ts
```diff
@@ -44,6 +45,42 @@ describe('computeToLock', () => {
       ),
     ).toEqual(10670n);
   });
+
+  it('should use the locked price if lower', () => {
```
turadg, line 49: :clap:

src/utils/vaultMath.ts
samsiegart: In zoe? I guess, it's small enough and only used in the UI though.

src/utils/hooks.ts
samsiegart: Yea it was originally showing negative and it seemed awkward. When it's at 0 it means it's about to disappear next block update after liquidation starts anyway so a very transient state.

I think I'll keep the impl because it takes long to run through the test scenario again and I don't think the suggestion will work without tweaks.

src/utils/vaultMath.ts
samsiegart: How about `quotePrice`? Because it's under `.quote` in vstorage.

src/utils/vaultMath.ts
turadg: Not trying to make more work :-) 

It illustrates the absence of max/min in ratio math. It's simple but also something one can expect to be provided by a math lib. This came up for TimeMath recently.

turadg (approved):

src/utils/vaultMath.ts
samsiegart: Ah I see, yea I wasn't sure if this would be useful more generally with the optional `priceB?`

PR #123, opened 2023-05-23
feat: enforce and warn about vault limit

### LINKED ISSUE #98: limit the number of vaults per account

The UI can only manage so many vaults.

## Description of the Design

When the account has 100 vaults, put a notice that at 150 it won't create any more. 

When it has 150, disable vault creation. Put a notice that to create more the user can provision a new wallet.

samsiegart:
fixes #98 

### Warning:

warning

### Limit Reached:

image

### DIFF
--- a/src/components/NewVaultOfferSummary.tsx
+++ b/src/components/NewVaultOfferSummary.tsx
@@ -28,6 +28,9 @@ const TableRow = ({ left, right }: TableRowProps) => {
   );
 };
 
+const vaultLimit = 150;
+const vaultLimitWarningThreshold = 100;
+
 const NewVaultOfferSummary = () => {
   const { collateralizationRatioError, toLockError, toReceiveError } =
     useAtomValue(inputErrorsAtom);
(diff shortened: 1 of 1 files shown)


src/components/NewVaultOfferSummary.tsx
```diff
@@ -110,30 +113,44 @@ const NewVaultOfferSummary = () => {
   const hasErrors =
     collateralizationRatioError || toLockError || toReceiveError;
 
+  const vaultLimitReached = (userVaults?.size ?? 0) >= 150;
```
turadg, line 116: `(userVaults?.size ?? 0)` is used twice. please factor it out to something like `userVaultCount`.

please make the magic numbers 150 and 100 into config constants.

src/components/NewVaultOfferSummary.tsx
```diff
@@ -110,30 +113,44 @@ const NewVaultOfferSummary = () => {
...
   const createVault = async () => {
     await makeOpenVaultOffer(depositAmount, mintAmount);
     setIsVaultCreationDialogOpen(true);
   };
 
+  const vaultLimitWarning =
+    (userVaults?.size ?? 0) >= 100 ? (
+      <div className="my-2 text-interOrange font-serif text-xs">
+        This account has created {userVaults?.size} vaults. There is a limit of
+        150 vaults per account.
```
turadg, line 137: the config const should appear here

samsiegart: Done

turadg (approved):

PR #124, opened 2023-05-23
Default to manage vaults view if user has created vault

### LINKED ISSUE #82: Default view for vault holder after wallet connection is 'Create a Vault' not 'My Vaults'
Currently after a user creates a vault, when they relaunch Vaults UI and connect their wallet, their default view is 'Create a Vault'. Once we detect they have a vault opened, their default view should be 'My Vaults' as user is likely to see vault overview rather than want to open a new vault upon returning.

## Description of the Design

Detect if user has open vault upon wallet connection. If yes -> take user to 'My Vaults' view. If No -> default view is 'Create Vault' view.

## Security Considerations
Unknown

## Test Plan
Will be smoke-tested

samsiegart:
fixes #82 

- Automatically try to connect wallet on load if user has connected before (stored in local storage).
- Default to "manage vaults" view if user has created a vault before (stored in local storage).
- Added loading animations to smoothen the "manage vaults" view.

### DIFF
--- a/src/components/ConnectWalletButton.tsx
+++ b/src/components/ConnectWalletButton.tsx
@@ -5,10 +5,13 @@ import {
   ChainConnection,
   chainConnectionAtom,
   isWalletConnectionInProgressAtom,
+  localStorageStore,
   networkConfigAtom,
   walletServiceAtom,
 } from 'store/app';
 import clsx from 'clsx';
+import { useEffect } from 'react';
+import { useStore } from 'zustand';
 
 const truncatedAddress = (chainConnection: ChainConnection) =>
   chainConnection.address.substring(chainConnection.address.length - 7);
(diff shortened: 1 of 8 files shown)


src/components/ConnectWalletButton.tsx
```diff
@@ -18,6 +21,23 @@ const ConnectWalletButton = () => {
   const isConnectionInProgress = useAtomValue(isWalletConnectionInProgressAtom);
   const chainConnection = useAtomValue(chainConnectionAtom);
   const { url } = useAtomValue(networkConfigAtom);
+  const { hasWalletPreviouslyConnected } = useStore(localStorageStore);
+
+  // Automatically connect if the user has previously connected.
```
turadg, line 26: nice

turadg (approved):

Issue #106, opened 2023-04-29
Adjust Vault usability enhancements batch #2


otoole-brendan:
## What is the Problem Being Solved?
Identified usability enhancements for 'adjust vault view'
## Description of the Design

- [ ]  Edit dropdown text from 'Action' -> 'Select Action' for both Debt and Collateral

- [ ]  Add instructions tile under Vault Summary

Most of these are shown in design here: https://www.figma.com/file/r0WcCjyS0LAOalFGV4DSuk/agoric-v1?node-id=2838-9618&t=I3FyVJiunOkVG8lV-0

Issue #83, opened 2023-03-09
When my vault is at risk of liquidation, it's no longer obvious when I click into the vault to adjust


otoole-brendan:
## What is the Problem Being Solved?
When my vault is at risk of liquidation a warning is shown on 'My Vaults' view but when I click the vault to adjust and take action, the warning does not persist. It should persist to keep informing the user that they need to take action to avoid liquidation. The lack of warning down-plays the severity of their situation.

## Description of the Design

Add 'Vault is at risk. Please increase your collateral or repay your outstanding IST debt to avoid liquidation.' to adjust vault view for affected vault (same red text/font as on my vaults view)

Warning in My Vaults when vault is at risk:
![image](https://user-images.githubusercontent.com/93355546/224111015-d4989e6f-9312-4bc5-a730-c17d3d9033bd.png)
Once I click into the vault to adjust, warning does not persist.
![image](https://user-images.githubusercontent.com/93355546/224111049-ed3e7f4c-1357-4f0a-a422-fd84bfbd5cb6.png)

## Security Considerations
N/A
## Test Plan
Will be smoke-tested

Issue #46, opened 2023-01-31
Consolidate styles into theme


samsiegart:
## What is the Problem Being Solved?

The dapp makes use of custom tailwind classes frequently e.g. `text-[#A3A5B9]`. Some of these are used once, but several are used in multiple places. These should be named and incorporated into the theme so they can be reused easily.

There are also many values that are close but not completely the same, we should consider making these conform together to reduce complexity.

PR #126, opened 2023-05-25
Adjust Vault view enhancements

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/106
fixes https://github.com/Agoric/dapp-inter/issues/83

### New "Instructions" Panel:

image

### DIFF
--- a/src/components/AdjustVault.tsx
+++ b/src/components/AdjustVault.tsx
@@ -44,6 +44,7 @@ const AdjustVault = () => {
     totalLockedValue,
     totalDebt,
     lockedPrice,
+    isAtRisk,
   } = vaultToAdjust;
 
   const [netVaultValue, isNetValueNegative] = netValue(
@@ -58,6 +59,12 @@ const AdjustVault = () => {
 
   return (
     <>
+      {isAtRisk && (
+        <div className="text-[#E22951] text-lg mt-4 font-serif font-medium">
+          This vault is at risk. Please increase your collateral or repay your
+          outstanding IST debt to avoid liquidation.
+        </div>
+      )}
       <div className="w-full flex justify-between mt-6 flex-wrap">
         <div className="font-serif flex items-baseline gap-3">
           <div className="font-medium text-2xl">{vaultLabel}</div>
@@ -66,7 +73,7 @@ const AdjustVault = () => {
         <div className="flex gap-8">
           <div>
             Current Price:{' '}
-            <span className="text-[#00B1A6] font-medium text-lg">
+            <span className="font-medium text-lg">
               {displayPrice(collateralPrice)}
             </span>
           </div>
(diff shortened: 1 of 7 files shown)


src/components/AdjustVault.tsx
```diff
@@ -58,6 +60,12 @@ const AdjustVault = () => {
 
   return (
     <>
+      {isAtRisk && (
+        <div className="text-[#E22951] text-lg mt-4 font-serif font-medium">
+          This vault is at risk. Please increase your collateral or repay your
```
turadg, line 65: "Please" doesn't sound right here. Perhaps "You can". 

Non-blocking as this is a Product matter and trivial to change.

src/components/AdjustVaultInstructions.tsx
```diff
@@ -0,0 +1,18 @@
+export const AdjustVaultInstructions = () => (
+  <div className="bg-white shadow-[0_22px_34px_rgba(116,116,116,0.25)] rounded-[10px] px-7 py-4">
+    <h3 className="text-[#484848] font-bold font-serif">Instructions</h3>
+    <ol className="list-decimal text-[#A3A5B9] text-sm mx-4 my-3 [&>li]:mt-1">
```
turadg, line 4: 20 other`A3A5B9` in main. Time to name this one?

src/components/AdjustVaultInstructions.tsx
```diff
@@ -0,0 +1,18 @@
+export const AdjustVaultInstructions = () => (
+  <div className="bg-white shadow-[0_22px_34px_rgba(116,116,116,0.25)] rounded-[10px] px-7 py-4">
+    <h3 className="text-[#484848] font-bold font-serif">Instructions</h3>
```
turadg, line 3: the only other use of `484848` in the app palette is a drop shadow. Is there another gray we've used for this text case?

src/utils/vaultMath.ts
```diff
@@ -7,8 +7,18 @@ import {
...
 } from '@agoric/zoe/src/contractSupport';
 import { AmountMath } from '@agoric/ertp';
-import { DebtSnapshot, PriceDescription, Ratio } from 'store/vaults';
-import { Amount, NatValue } from '@agoric/ertp/src/types';
+import {
+  DebtSnapshot,
+  LiquidationAuctionBook,
+  LiquidationSchedule,
+  PriceDescription,
+  PriceQuote,
```
turadg, line 15: lint

src/components/AdjustVaultInstructions.tsx
samsiegart: I'd like to do that as part of https://github.com/Agoric/dapp-inter/issues/46, which is second in my queue currently behind https://github.com/Agoric/dapp-inter/issues/127

src/components/AdjustVaultInstructions.tsx
samsiegart: Good point. It has a variable already `mineShaft`, but it's actually the default text color, so just removed this.

src/utils/vaultMath.ts
samsiegart: Removed thanks

src/components/AdjustVault.tsx
otoole-brendan: prefer to keep it there - we want users to avoid liquidation so I'm in favor of using strong language here. 'You can' doesn't imply urgency the same way as 'Please..

PR #129, opened 2023-05-25
feat: auto-dismiss no smart wallet toast

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/6552

### DIFF
--- a/src/service/wallet.ts
+++ b/src/service/wallet.ts
@@ -68,6 +68,8 @@ type ConnectionError = {
   message: string;
 };
 
+const autoCloseDelayMs = 7000;
+
 export const makeWalletService = () => {
   let stopWatchingPurses: () => void;
   let stopWatchingPublicSubscribers: () => void;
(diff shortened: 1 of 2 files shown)


src/service/wallet.ts
```diff
@@ -119,14 +119,19 @@ export const makeWalletService = () => {
...
           break;
         case Errors.networkConfig:
           showToast('Network not found.');
           break;
         case Errors.noSmartWallet:
-          showToast(React.createElement(SmartWalletNotFoundToast, null));
+          showToast(
+            React.createElement(SmartWalletNotFoundToast, {
+              hideProgressBar: false,
+              autoClose: 7000,
```
turadg, line 132: are these auto-close values to remain the same? if so please make into a constant.

src/service/wallet.ts
```diff
@@ -119,14 +119,19 @@ export const makeWalletService = () => {
...
+            autoClose: 7000,
           });
           break;
         case Errors.networkConfig:
           showToast('Network not found.');
           break;
         case Errors.noSmartWallet:
-          showToast(React.createElement(SmartWalletNotFoundToast, null));
+          showToast(
+            React.createElement(SmartWalletNotFoundToast, {
```
turadg, line 130: what would it take to make the message disappear once the smart wallet is made?

samsiegart: Also I'd like to move towards the end goal of an extension subsuming wallet.agoric.app (metamask snap, frontier etc.) sooner than later, and this is tangential to that.

PR #131, opened 2023-05-26
Consolidate styles

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/46

- Cleaned up the theme variables, removing bespoke styles (`[]` in `className`s) where possible.
- Used correct format for theme colors
- Changed some values that were close enough to barely be noticeable so they could reuse the same colors/shadows etc.
- Created some custom classes for buttons and stuff

The only `[]` remaining in the `className`s  as far as I can tell are one-offs, or have to do with spacing/layout.


turadg (approved):
Awesome

PR #135, opened 2023-05-31
feat: batch chain storage query library

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/6870

This is in its own directory so that it's more portable. It will be copy-pasted into other dapps, but eventually live in a reusable package.

## Testing

Tested manually after rewiring dapp to use the client, as well as unit tests. NOTE - The wallet connection component in agoric-sdk doesn't use this lib yet, so if you connect your wallet then you'll see a lot of unnecessary rpc requests still.

### DIFF
--- a/src/rpc/src/batchQuery.ts
+++ b/src/rpc/src/batchQuery.ts
@@ -32,7 +32,7 @@ export const batchVstorageQuery = (
             );
           }
 
-          const data = JSON.parse(atob(entry.result.response.value));
+          const data = JSON.parse(window.atob(entry.result.response.value));
 
           if (paths[index][0] === AgoricChainStoragePathKind.Children) {
             return [
@@ -43,11 +43,12 @@ export const batchVstorageQuery = (
 
           const value = JSON.parse(data.value);
 
+          const latestValueStr = value.values[value.values.length - 1];
           return [
             JSON.stringify(paths[index]),
             {
               blockHeight: value.blockHeight,
-              value: unserialize(JSON.parse(value.values.pop())),
+              value: unserialize(JSON.parse(latestValueStr)),
             },
           ];
         }),
--- a/src/rpc/src/batchQuery.ts
+++ b/src/rpc/src/batchQuery.ts
@@ -23,13 +23,24 @@ export const batchVstorageQuery = (
     .then(res =>
       Object.fromEntries(
         (Array.isArray(res) ? res : [res]).map((entry, index) => {
+          if (entry.result.response.code) {
+            return [
+              JSON.stringify(paths[index]),
+              { error: entry.result.response.log },
+            ];
+          }
+
           if (!entry.result.response.value) {
-            throw new Error(
-              'Cannot parse value of response for path' +
-                paths[index] +
-                ': ' +
-                entry,
-            );
+            return [
+              JSON.stringify(paths[index]),
+              {
+                error:
+                  'Cannot parse value of response for path [' +
+                  paths[index] +
+                  ']: ' +
+                  JSON.stringify(entry),
+              },
+            ];
           }
 
           const data = JSON.parse(window.atob(entry.result.response.value));
--- a/src/rpc/src/batchQuery.ts
+++ b/src/rpc/src/batchQuery.ts
@@ -1,0 +1,56 @@
+import { AgoricChainStoragePathKind } from './types';
+import type { Unserialize } from '@endo/marshal';
+
+export const batchVstorageQuery = (
+  node: string,
+  unserialize: Unserialize<string>,
+  paths: [AgoricChainStoragePathKind, string][],
+) => {
+  const options = {
+    method: 'POST',
+    body: JSON.stringify(
+      paths.map(path => ({
+        jsonrpc: '2.0',
+        id: 1,
+        method: 'abci_query',
+        params: { path: `/custom/vstorage/${path[0]}/${path[1]}` },
+      })),
+    ),
+  };
+
+  return fetch(node, options)
+    .then(res => res.json())
+    .then(res =>
+      Object.fromEntries(
+        (Array.isArray(res) ? res : [res]).map((entry, index) => {
+          if (!entry.result.response.value) {
+            throw new Error(
+              'Cannot parse value of response for path' +
+                paths[index] +
+                ': ' +
+                entry,
+            );
+          }
+
+          const data = JSON.parse(atob(entry.result.response.value));
+
+          if (paths[index][0] === AgoricChainStoragePathKind.Children) {
+            return [
+              JSON.stringify(paths[index]),
+              { value: data.children, blockHeight: undefined },
+            ];
+          }
+
+          const value = JSON.parse(data.value);
+
+          return [
+            JSON.stringify(paths[index]),
+            {
+              blockHeight: value.blockHeight,
+              value: unserialize(JSON.parse(value.values.pop())),
+            },
+          ];
+        }),
+      ),
+    );
+};
--- a/src/rpc/src/batchQuery.ts
+++ b/src/rpc/src/batchQuery.ts
@@ -1,6 +1,17 @@
 import { AgoricChainStoragePathKind } from './types';
 import type { Unserialize } from '@endo/marshal';
 
+export const pathToKey = (path: [AgoricChainStoragePathKind, string]) =>
+  path.join('.');
+
+export const keyToPath = (key: string) => {
+  const parts = key.split('.');
+  return [parts[0], parts.slice(1).join('.')] as [
+    AgoricChainStoragePathKind,
+    string,
+  ];
+};
+
 export const batchVstorageQuery = (
   node: string,
   unserialize: Unserialize<string>,
@@ -25,14 +36,14 @@ export const batchVstorageQuery = (
         (Array.isArray(res) ? res : [res]).map((entry, index) => {
           if (entry.result.response.code) {
             return [
-              JSON.stringify(paths[index]),
+              pathToKey(paths[index]),
               { error: entry.result.response.log },
             ];
           }
 
           if (!entry.result.response.value) {
             return [
-              JSON.stringify(paths[index]),
+              pathToKey(paths[index]),
               {
                 error:
                   'Cannot parse value of response for path [' +
@@ -47,7 +58,7 @@ export const batchVstorageQuery = (
 
           if (paths[index][0] === AgoricChainStoragePathKind.Children) {
             return [
-              JSON.stringify(paths[index]),
+              pathToKey(paths[index]),
               { value: data.children, blockHeight: undefined },
             ];
           }
@@ -56,7 +67,7 @@ export const batchVstorageQuery = (
 
           const latestValueStr = value.values[value.values.length - 1];
           return [
-            JSON.stringify(paths[index]),
+            pathToKey(paths[index]),
             {
               blockHeight: value.blockHeight,
               value: unserialize(JSON.parse(latestValueStr)),
--- a/src/rpc/src/chainStorageWatcher.ts
+++ b/src/rpc/src/chainStorageWatcher.ts
@@ -1,5 +1,5 @@
 import { AgoricChainStoragePathKind } from './types';
-import { batchVstorageQuery } from './batchQuery';
+import { batchVstorageQuery, keyToPath, pathToKey } from './batchQuery';
 import type { UpdateHandler } from './types';
 import type { Unserialize } from '@endo/marshal';
 
@@ -86,9 +86,7 @@ export const makeAgoricChainStorageWatcher = (
     nextQueryTimeout = null;
     isNewPathWatched = false;
 
-    const paths = [...watchedPathsToSubscribers.keys()].map(pathKey =>
-      JSON.parse(pathKey),
-    );
+    const paths = [...watchedPathsToSubscribers.keys()].map(keyToPath);
 
     if (!paths.length) {
       isQueryInProgress = false;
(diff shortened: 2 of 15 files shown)


src/rpc/test/chainStorageWatcher.test.ts
```diff
@@ -0,0 +1,235 @@
...
+        },
+      ]),
+    );
+
+    vi.advanceTimersToNextTimer();
+    expect(await values[1].value).toEqual(expected2);
+    expect(fetch).toHaveBeenCalledTimes(2);
+  });
+
+  it('notifies for changed children values', async () => {
```
samsiegart, line 115: I also tried writing a test to make sure it *doesn't* notify for *unchanged* values, but struggled with the fake timers for a long time, unable to find a sufficient workaround.

vitest.config.ts
```diff
@@ -6,7 +6,6 @@ export default mergeConfig(
   viteConfig,
   defineConfig({
     test: {
-      setupFiles: ['src/installSesLockdown.ts'],
```
samsiegart, line 9: The fake timers don't work with lockdown, so I moved this out of the global setup.

src/rpc/src/batchQuery.ts
```diff
@@ -0,0 +1,56 @@
...
+          if (!entry.result.response.value) {
+            throw new Error(
+              'Cannot parse value of response for path' +
+                paths[index] +
+                ': ' +
+                entry,
+            );
+          }
+
+          const data = JSON.parse(atob(entry.result.response.value));
```
turadg, line 35: running in Node this will import Node's `atob` which is deprecated. To clarify and avoid that warning please use,
```suggestion
          const data = JSON.parse(window.atob(entry.result.response.value));
```

src/rpc/src/batchQuery.ts
```diff
@@ -0,0 +1,56 @@
...
+
+          const value = JSON.parse(data.value);
+
+          return [
+            JSON.stringify(paths[index]),
+            {
+              blockHeight: value.blockHeight,
+              value: unserialize(JSON.parse(value.values.pop())),
+            },
+          ];
```
turadg, line 52: `pop()` mutates the array, which is confusing here because the array is never read again.

Let's also clarify why we're taking the end of the array:
```suggestion
          const latestValueStr = value.values[value.values.length - 1];
          return [
            JSON.stringify(paths[index]),
            {
              blockHeight: value.blockHeight,
              value: unserialize(JSON.parse(latestValueStr)),
            },
          ];
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+import type { UpdateHandler } from './types';
+import type { Unserialize } from '@endo/marshal';
+
+type Subscriber<T> = {
+  onUpdate: UpdateHandler<T>;
+};
+
+const defaultNewPathQueryDelayMs = 20;
+const defaultRefreshLowerBoundMs = 2000;
+const defaultRefreshUpperBoundMs = 4000;
```
turadg, line 12: glad to see these constants. consider the following to match the strings of the params:
```suggestion
const defaults = {
  newPathQueryDelayMs: 20,
  refreshLowerBoundMs: 2000,
  refreshUpperBoundMs: 4000,
}
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+
+const randomRefreshPeriod = (
+  refreshLowerBoundMs: number,
+  refreshUpperBoundMs: number,
+) =>
+  Math.floor(Math.random() * (refreshUpperBoundMs - refreshLowerBoundMs)) +
```
turadg, line 18: floor seems oddly specific. wouldn't `round` work?

also curious: does something fail if this returns a floating point?

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+  let isQueryInProgress = false;
+  let nextQueryTimeout: number | null = null;
+
+  const queueNextQuery = () => {
+    if (isQueryInProgress || !watchedPathsToSubscribers.size) {
+      return;
+    }
+
+    if (isNewPathWatched) {
+      // If there's new paths to watch, schedule another query very soon.
```
turadg, line 59:
```suggestion
      // If there are new paths to watch, schedule another query very soon.
```
or
```suggestion
      // If there is any new path to watch, schedule another query very soon.
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        latestValueCache.set(path, [
+          blockHeight ?? JSON.stringify(value),
+          value,
+        ]);
+
+        subscribers.forEach(s => {
+          s.onUpdate(harden(value));
+        });
```
turadg, line 109: prefer control structures.
```suggestion
        for (const s of subscribers) {
          s.onUpdate(harden(value));
        }
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+    );
+
+    if (!paths.length) {
+      isQueryInProgress = false;
+      return;
+    }
+
+    try {
+      const data = await batchVstorageQuery(rpcAddr, unserialize, paths);
+      watchedPathsToSubscribers.forEach((subscribers, path) => {
```
turadg, line 89: prefer control structures
```suggestion
      for (const [path, subscribers] of watchedPathsToSubscribers.entries()) => {
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+    } finally {
+      isQueryInProgress = false;
+      queueNextQuery();
+    }
+  };
+
+  const stopWatching = (pathKey: string, subscriber: Subscriber<unknown>) => {
+    const subscribersForPath = watchedPathsToSubscribers.get(pathKey);
+    if (!subscribersForPath) {
+      throw new Error('Trying to unsubscribed from unwatched path');
```
turadg, line 120:
```suggestion
      throw new Error(`cannot unsubscribe from unwatched path ${pathKey}`);
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    if (subscribersForPath.size === 1) {
```
turadg, line 123: this code prevents `subscribersForPath.size === 0` but should be defensive against that bug. e.g. `assert(subscribersForPath.size >= 1);`

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+      isNewPathWatched = true;
+      queueNextQuery();
+    }
+  };
+
+  const watchLatest = <T>(
+    path: [AgoricChainStoragePathKind, string],
+    onUpdate: (latestValue: T) => void,
+  ) => {
+    const pathKey = JSON.stringify(path);
```
turadg, line 142: to make pathKey easier to read in debugging and error messages, WDYT of making the same vstorage key format that we use elsewhere?
```suggestion
    const pathKey = path.join(',');
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+  refreshLowerBoundMs = defaultRefreshLowerBoundMs,
+  refreshUpperBoundMs = defaultRefreshUpperBoundMs,
+) => {
+  // Map of paths to [identifier, value] pairs of most recent response values.
+  //
+  // The 'identifier' is used to avoid notifying subscribers for already-seen
+  // values. For 'data' queries, 'identifier' is the blockheight of the
+  // response. For 'children' queries, 'identifier' is the stringified array
+  // of children.
+  const latestValueCache = new Map<string, [string, unknown]>();
```
turadg, line 46: the elements of the tuple can be named
```suggestion
  const latestValueCache = new Map<
    string,
    [identifier: string, value: unknown]
  >();
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+  const watchLatest = <T>(
+    path: [AgoricChainStoragePathKind, string],
+    onUpdate: (latestValue: T) => void,
+  ) => {
+    const pathKey = JSON.stringify(path);
+    const subscriber = makePathSubscriber(onUpdate);
+
+    const latestValue = latestValueCache.get(pathKey);
+    if (latestValue) {
+      (subscriber as Subscriber<unknown>).onUpdate(harden(latestValue[1]));
```
turadg, line 147: a little cleaner
```suggestion
      subscriber.onUpdate(harden(latestValue[1]) as T);
```

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+) => {
+  // Map of paths to [identifier, value] pairs of most recent response values.
+  //
+  // The 'identifier' is used to avoid notifying subscribers for already-seen
+  // values. For 'data' queries, 'identifier' is the blockheight of the
+  // response. For 'children' queries, 'identifier' is the stringified array
+  // of children.
+  const latestValueCache = new Map<string, [string, unknown]>();
+
+  const watchedPathsToSubscribers = new Map<string, Set<Subscriber<unknown>>>();
```
turadg, line 48: nit: each Subscriber is a different type, depending on the string key. so this isn't really a JS Map of the same value with arbitrary keys. It's more of an map-as-object mapping specific keys to specific values.

If you made it an object you could have a type for it that mapped each vstorage key to its subscriber type. E.g.
```
const watchedPathsToSubscribers = {} as {
  'vaultFactory.metrics'?: Set<Subscriber<VaultFactoryMetricsNotification>>,
};
```
Since that would get repetitive, there could be a named type like, 
```
type vstorageSchema = {
 'vaultFactory.metrics': VaultFactoryMetricsNotification;
}
```
And then this mapping could be generated from that type.

Not at all a blocker :)  But consider factoring this now as an object instead of a map, so that adopting this idea would not require runtime changes.

src/rpc/src/chainStorageWatcher.ts
```diff
@@ -0,0 +1,167 @@
...
+    const subscriber = makePathSubscriber(onUpdate);
+
+    const latestValue = latestValueCache.get(pathKey);
+    if (latestValue) {
+      (subscriber as Subscriber<unknown>).onUpdate(harden(latestValue[1]));
+    }
+
+    const samePathSubscribers = watchedPathsToSubscribers.get(pathKey);
+    if (samePathSubscribers !== undefined) {
+      samePathSubscribers.add(subscriber as Subscriber<unknown>);
```
turadg, line 152: these three casts to unknown would go away if the map were instead a fully typed object

src/rpc/test/chainStorageWatcher.test.ts
```diff
@@ -0,0 +1,235 @@
...
+const fakeRpcAddr = 'https://agoric-rpc.vitest-nodes.com:443';
+const serialize = (val: unknown) => val;
+const unserialize = (val: any) => val;
+
+let watcher: ReturnType<typeof makeAgoricChainStorageWatcher>;
+
+describe('makeAgoricChainStorageWatcher', () => {
+  beforeEach(() => {
+    watcher = makeAgoricChainStorageWatcher(fakeRpcAddr, unserialize);
+    vi.useFakeTimers();
```
turadg, line 18: does this need to be before *each* or could it be before all? everything in this file uses fake timers. does vitest leak that option to other tests? (iow what would happen if `vi.useRealTimers()` were omitted?)

turadg (commented):
> This PR just adds the library, actually wiring it up inside the dapp will come later.

It's good that this library has its own commit. To really know it works the test plan should have it tested in dapp-inter. I see you tested it manually there but as the reviewer I'd like to test it to.

Please include in this PR using the new library. (Indeed, unless it's used, it's not a feature. dapp-inter doesn't export an API)

src/rpc/test/chainStorageWatcher.test.ts
samsiegart: I'm not sure, does it matter? I'm just following the docs https://vitest.dev/guide/mocking.html#dates

test/utils/vaultMath.test.ts
```diff
@@ -1,3 +1,4 @@
+import '../../src/installSesLockdown';
 import 'ses'; // adds harden
```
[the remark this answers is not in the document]
samsiegart: Removed

turadg (approved):

samsiegart:
@turadg I see that you approved this already, but I just updated the PR to include migrating the dapp to use this library, removing casting as a (direct) dependency entirely. PTAL.

src/rpc/src/chainStorageWatcher.ts
turadg: If there weren't a fully unambiguous way to split the string then the vstorage system wouldn't work. 

The constraint allowing join and split by '.' is that path segments can't contain that character.

src/rpc/test/chainStorageWatcher.test.ts
turadg: It matters to understanding Vitest execution model. I'm fine to cargo cult the docs for now.

src/service/vaults.ts
```diff
@@ -19,46 +17,40 @@ import type {
   VaultMetrics,
   LiquidationAuctionBook,
 } from 'store/vaults';
+import { AgoricChainStoragePathKind as Kind } from 'rpc';
 
 type ValuePossessor<T> = {
```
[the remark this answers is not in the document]
samsiegart: Not for the top-level data structures, but some data has fields like `EndorsedUI` where it has a `value` field for its actual value.

PR #137, opened 2023-06-06
bump deps

turadg:
- ci: match local dev to CI node-version
- chore(deps): bump happy-dom to 9.20.3
- chore(deps): bump @endo/* to latest
- chore(deps): bump Vite to v4
- chore(deps): npx update-browserslist-db@latest
- build: loosen chunkSizeWarningLimit
- chore(deps): bump TypeScript to 5.1
- chore(deps): bump @vitejs/plugin-react to v4
- chore(deps): rm unused testing-library/react
- chore(deps): bump vite-tsconfig-paths to 4.2.0
- chore(deps): yarn-deduplicate


patches/vitest+0.32.0.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/node_modules/vitest/dist/vendor-rpc.4d3d7a54.js b/node_modules/vitest/dist/vendor-rpc.4d3d7a54.js
+index c98baeb..e365e91 100644
+--- a/node_modules/vitest/dist/vendor-rpc.4d3d7a54.js
++++ b/node_modules/vitest/dist/vendor-rpc.4d3d7a54.js
+@@ -2,19 +2,16 @@ import { getSafeTimers } from '@vitest/utils';
+ import { g as getWorkerState } from './vendor-global.6795f91f.js';
+ 
+ const { get } = Reflect;
+-const safeRandom = Math.random;
```
samsiegart, line 9: It's not clear why this is needed or if there's unintended consequences. If those are known, a comment would be good. Otherwise, I think it's less critical to keep our testing framework as up-to-date as runtime dependencies, and maybe the overhead of maintaining a patch isn't worth it. Otherwise LGTM though.

samsiegart (approved):

patches/vitest+0.32.0.patch
turadg: Good point. I added a comment: [866c981](https://github.com/Agoric/dapp-inter/pull/137/commits/866c981f8f6bc7ad33e605fad6bfc592bd1f9c39)

PR #139, opened 2023-06-07
Renames

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/7588

Backward-compatible renames in preparation for https://github.com/Agoric/agoric-sdk/pull/7871

## Testing

Tested with ollinet and against a local chain running the above PR of sdk

### DIFF
--- a/src/service/vaults.ts
+++ b/src/service/vaults.ts
@@ -215,6 +215,8 @@ const watchUserVaults = () => {
 
 type GoverenedParamsCurrent = {
   DebtLimit: ValuePossessor<Amount<'nat'>>;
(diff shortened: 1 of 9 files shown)


src/service/vaults.ts
```diff
@@ -253,7 +270,8 @@ export const watchVaultFactory = () => {
       value => {
         console.debug('got update', path, value);
         const { current } = value;
-        const interestRate = current.InterestRate.value;
+        const stabilityFee =
+          current.StabilityFee?.value ?? value.current.InterestRate?.value;
```
turadg, line 274: good isolation

turadg (approved):

PR #143, opened 2023-06-20
fix: rely on ids not ordering for batch query

samsiegart:
@gibson042 Pointed out that the ordering of responses is not guaranteed in batch queries, we should rely on the id field instead


turadg (approved):
LGTM.

Time to put this into ui-kit so changes don't need to be copied across repos?

samsiegart:
> Time to put this into ui-kit so changes don't need to be copied across repos?

It's only in one repo so far, but yea I suppose I'll do that next

PR #144, opened 2023-06-22
build: use npm package for rpc

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/6870


turadg (approved):
LGTM. Thanks for care in the conventional commit message.

PR #147, opened 2023-06-24
fix: connect wallet button bugs

samsiegart:
- Don't let the user click connect until `chainStorageWatcher` is initialized (which happens after network config is fetched). That's because `makeAgoricWalletConnection` requires it.
- Fix the loading spinner animation. We changed the css variables and that broke the styles there.

### DIFF
--- a/src/components/ConnectWalletButton.tsx
+++ b/src/components/ConnectWalletButton.tsx
@@ -53,13 +53,16 @@ const ConnectWalletButton = () => {
     return 'Connect Wallet';
   })();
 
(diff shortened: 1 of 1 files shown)


turadg (approved):
What would it take to have regression tests for these?

samsiegart:
> What would it take to have regression tests for these?

For the style fix, I think that might be tricky. Possibly some type of screenshot testing, but with an animation like that it might not be reliable to test.

For the button thing, I think a component test with vitest and mocking the store would be doable

PR #190, opened 2023-07-10
build: update web-components to remove patch

samsiegart:


turadg (approved):
Worth publishing a real release to depend on here?

samsiegart:
> Worth publishing a real release to depend on here?

Will do so but won't block this PR on it.

Issue #159, opened 2023-06-27
Show locked liquidation price when known


otoole-brendan:
## What is the Problem Being Solved?
IOU - ref https://agoricopco.slack.com/archives/C03FUCJV2J2/p1684169603510209
## Description of the Design
IOU

PR #192, opened 2023-07-12
fix: use lockedQuote for auction price

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/159

Functionality affected by https://github.com/Agoric/agoric-sdk/issues/8041 but not a blocker

## Screenshots

liquidationcapture

capture2

capture3

### DIFF
--- a/src/service/vaults.ts
+++ b/src/service/vaults.ts
@@ -14,9 +14,9 @@ import type {
   LiquidationSchedule,
   VaultInfoChainData,
   VaultMetrics,
-  LiquidationAuctionBook,
 } from 'store/vaults';
 import { AgoricChainStoragePathKind as Kind } from '@agoric/rpc';
+import { invertRatio } from '@agoric/zoe/src/contractSupport';
 
 type ValuePossessor<T> = {
   value: T;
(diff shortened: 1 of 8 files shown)


src/service/vaults.ts
```diff
@@ -308,7 +306,11 @@ export const watchVaultFactory = () => {
       [Kind.Data, path],
       value => {
         console.debug('got update', path, value);
-        useVaultStore.getState().setVaultMetrics(id, value);
+        useVaultStore.getState().setVaultMetrics(id, {
+          ...value,
+          // Invert the lockedQuote to be congruent with oracle price quotes.
+          lockedQuote: value.lockedQuote && invertRatio(value.lockedQuote),
```
turadg, line 312: are there distinct names that would make sense in this codebase? for when someone is reading the symbol out of context.

turadg (commented):
Glad to see the red :) 

I left a question. Not a blocker but I'm going to leave Approval to @iomekam .

src/service/vaults.ts
samsiegart: I'm not sure if I can think of a better name, it might be easiest to just keep the on-chain name here.

PR #195, opened 2023-08-15
feat: rough demo of keplr signing

samsiegart:
The UX is unrefined, demo purposes only right now, but it shows the capabilities off.

Only works when built locally with:
`yarn link @agoric/web-components`
`yarn link @agoric/rpc`


turadg:
Even just reading the toasts it's clear the UX is better

Issue #201, opened 2023-09-15
Several users unable to use dapps due to read-only mode - unable to connect to wallet


otoole-brendan:
## Describe the bug

We have two users currently reporting dapp issues (@rowland and Anil of Bytepitch) who are unable to get past read-only mode as they're not able to connect the dapps to wallet.

They are unable to prompt dapp connection request from the dapp.

## To Reproduce

[IOU Rowland's info showing enabled

## Expected behavior

Dapp should connect to wallet and work

otoole-brendan:
Issue has been confirmed to be due to unpartitioned 3rd party storage trial on Chrome affected some users who have it enabled

PR #202, opened 2023-09-18
feat: remove wallet bridge

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/201


src/App.tsx
```diff
@@ -147,14 +135,13 @@ const App = () => {
...
       ></ToastContainer>
-      <OfferSignerBridge />
       <div className="w-screen max-w-7xl m-auto">
         {error ? (
           <>
             <div className="flex justify-end flex-row space-x-2 items-center mr-6 m-2">
               {networkDropdown}
             </div>
-            <div>Error connecting to chain</div>
+            <div>Error connecting to chain!</div>
```
turadg, line 144: lol

src/components/CloseVaultDialog.tsx
```diff
@@ -130,7 +120,7 @@ const CloseVaultDialog = ({
           {...noticeProps}
           className="overflow-hidden text-interOrange"
         >
-          Offer sent to Agoric Smart Wallet for approval.
+          Vault has been closed successfully.
```
turadg, line 123: It's only been sent, right? The transaction could still fail. For example if another transaction in flight took out more debt, this pay down to close would not be enough.

src/components/ManageVaults.tsx
```diff
@@ -111,7 +113,7 @@ const ManageVaults = () => {
     );
   } else if (!chainConnection) {
     popupContent = 'Connect your wallet to manage your vaults.';
-  } else if (vaults?.size === 0) {
+  } else if (vaults?.size === 0 || isSmartWalletProvisioned === false) {
```
turadg, line 116: Consider having its own case message. Perhaps "You have no Agoric smart wallet yet."

src/components/ProvisionSmartWalletDialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const provision = () => {
+    assert(chainConnection);
+    provisionSmartWallet(chainConnection, setProvisionToastId);
+    setShouldBeOpenIfFeeLoaded(false);
+  };
+
+  const body = (
+    <span>
+      To interact with contracts on the Agoric chain, a smart wallet must be
```
turadg, line 98: For future work: should we have a utility for this in ui-kit? Worth an issue imo

src/components/VaultAdjustmentDialog.tsx
```diff
@@ -9,30 +7,24 @@ type Props = {
   onClose: () => void;
 };
 
-const message =
-  'Your vault adjustment request has been successfully submitted. Go to your Agoric Smart Wallet to approve.';
+const message = "Your vault's balances have been updated.";
```
turadg, line 10: Same race issue as above.

src/components/VaultCreationDialog.tsx
```diff
@@ -9,30 +7,24 @@ type Props = {
...
-
   const goToVaults = () => {
     onClose();
     setViewMode(ViewMode.Manage);
   };
 
   return (
     <ActionsDialog
-      title="Success: Offer Submitted"
+      title="Success: Vault Created"
```
turadg, line 22: This may be okay sinxe the 'message' is accurate

src/service/vaults.ts
```diff
@@ -425,157 +426,148 @@ export const makeOpenVaultOffer = async (
...
+      }
+    },
+  );
 };
 
 type Proposal = {
   give: {
-    Collateral?: { amount: CapData<string> };
-    Minted?: { amount: CapData<string> };
+    Collateral?: Amount;
```
turadg, line 463: Consider adding 'nat'

turadg (commented):
Huge UX improvement. Want to make sure the messages are accurate before approving.

src/components/CloseVaultDialog.tsx
turadg: Oh just the variable name is off. Approving assuming you'll revise

turadg (approved):

src/service/vaults.ts

src/components/ProvisionSmartWalletDialog.tsx
samsiegart: Possibly for the provisioning fee, though I'm not sure how likely that flow is to change (i.e. if we implement automatic provisioning). We already have a utility for the provision status. I'm not sure if the whole dialog flow is worth making into a component depending on how dapps want to handle it

src/components/ManageVaults.tsx
samsiegart: I considered that but I'm leaning towards keeping this as is. We already show the dialog, so it's not necessary to tell them twice, and it doesn't matter to them that there's no smart wallet until they actually try to create a vault, where they'll see the "Provision" button in that view.

src/components/CloseVaultDialog.tsx
samsiegart: Updated the variable name

PR #216, opened 2023-12-19
feat: show more informative rpc error

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/8505
fixes https://github.com/Agoric/ui-kit/issues/57

This uses the new `@agoric/rpc` and `@agoric/web-components` to make some improvements to error handling as a precursor to the node selector. The new package versions have a few improvements:

- No longer use the batch vstorage API which is [deprecated](https://github.com/Agoric/agoric-sdk/blob/9380cfe04661eaed6de168e65a96730a8798bffb/golang/cosmos/x/vstorage/README.md#arbitrary-response-http-interface), instead poll each vstorage path from the API server separately
- No longer have a separate `onError` handler for each vstorage path. For missing data, it now just yields an empty `value`, and server errors are surfaced to the single `onError` handler of the `chainStorageWatcher`
- Add an error handler to wallet connection component for RPC/API errors which indicates which endpoint is failing (RPC or API)
- Automatically retry all queries before surfacing `onError` (in both wallet connection component and `chainStorageWatcher`)
- Fix some bugs in the wallet connection component around offer and smart-wallet-provision status updates

With all these changes, the dapp can now more gracefully and confidently handle server errors and show a more informative dialog instead of crashing with an unhelpful error message on the screen. In the future we can prompt the user to select a different node in this case, but for now just the dialog is being added.

## Screenshot
image


src/service/vaults.ts
```diff
@@ -155,6 +162,9 @@ const watchUserVaults = () => {
         [Kind.Data, path],
         value => {
           console.debug('got update', subscriber, value);
+          if (!value) {
+            return;
```
samsiegart, line 166: We might receive the update to `offerIdsToPublicSubscribers` from the user's wallet and start watching the vault before the vault data is actually published to vstorage, so we don't want to error out in that case. Just wait for the data to show up assuming it will eventually. I'm not sure how else to handle it, but worst case if the contract had a bug the vault would just perpetually show a loading state.

src/App.tsx
```diff
@@ -78,50 +82,56 @@ const useAppVersionWatcher = () => {
...
           }),
         );
 
         watchVbank();
       } catch (e) {
         if (isCancelled) return;
-        setError(e);
+        setChainConnectionError(
+          new Error(
+            `Error fetching network config from ${netConfig.url}` +
```
ivanlei, line 117: _question_:
Why is this using 
```
e instanceof Error ? `: ${e.message}` : ''
```
 while just above it's only  
```
e.message
```

Is it because we know `makeAgoricChainStorageWatcher` is giving us an instance of `Error` but we don't know if the `catch` is?

src/service/vbank.ts
```diff
@@ -16,30 +16,21 @@ export const watchVbank = () => {
...
-      appStore.setState({ brandToInfo });
-    },
-    log => {
-      console.error('Error watching vbank assets', log);
+  chainStorageWatcher.watchLatest<VbankUpdate>([Kind.Data, path], value => {
+    console.debug('got update', path, value);
+    if (!value) {
       appStore.setState({
-        watchVbankError: 'Error loading asset display info',
+        watchVbankError: `${path} returned undefined`,
```
gibson042, line 23:
```suggestion
        watchVbankError: `No value at ${path}`,
```
(better covers non-`undefined` falsy values in correspondence with the actual test)

src/service/wallet.ts
```diff
@@ -111,7 +112,20 @@ export const makeWalletService = () => {
...
-      const connection = await makeAgoricWalletConnection(chainStorageWatcher);
+      assert(rpcNode);
+      const connection = await makeAgoricWalletConnection(
+        chainStorageWatcher,
+        rpcNode,
+        (e: unknown) =>
+          appStore
+            .getState()
+            .setChainConnectionError(
+              new Error(
```
gibson042, line 123: Another good candidate for the suggested helper.

gibson042 (approved):
I'm not 100% clear on the React details, but close enough to feel confident in this review. No blocking issues spotted.

Issue #211, opened 2023-11-06
Add RPC Node selector (dapp-inter)


otoole-brendan:
## What is the Problem Being Solved?

We need to update rpc connection logic/design for dapp-inter to improve UX

## Description of the Design

Use node selector as per https://github.com/Agoric/agoric-sdk/issues/8505 for dapp-inter to improve RPC node experience

PR #217, opened 2023-12-27
feat: allow node selection after connection error

samsiegart:
refs https://github.com/Agoric/agoric-sdk/issues/8505
refs https://github.com/Agoric/dapp-inter/issues/211

This adds a "connection settings" button to the chain connection error dialog which allows you to select API and RPC nodes. It saves your selections to local storage, which take priority over whichever nodes are loaded from the network config.

It doesn't provide an entrypoint to the node selector outside of the error dialog, that can be added in a future PR. Also, if you change networks (dev/testnet only, as mainnet doesn't have the network dropdown), it just deletes the saved nodes, instead of saving a map of saved nodes for every network.

## Screenshots

image

image

## Recording

(Animations look choppy in the gif due to compression, but you can try it in the live demo in the comments yourself)
fixed-provision-bug

### DIFF
--- a/src/components/NodeSelectorDialog.tsx
+++ b/src/components/NodeSelectorDialog.tsx
@@ -1,0 +1,115 @@
+import {
+  chainStorageWatcherAtom,
+  isNodeSelectorOpenAtom,
+  networkConfigAtom,
+  rpcNodeAtom,
+  savedApiNodeAtom,
+  savedRpcNodeAtom,
+} from 'store/app';
+import { useAtom, useAtomValue, useSetAtom } from 'jotai';
+import ActionsDialog from './ActionsDialog';
+import Combobox from 'components/Combobox';
+import { useEffect, useRef, useState } from 'react';
+import { fetchAllAddrs } from 'utils/rpc';
+
+const useRpcAddrs = () => {
+  const [rpcAddrs, setRpcAddrs] = useState([]);
+  const [apiAddrs, setApiAddrs] = useState([]);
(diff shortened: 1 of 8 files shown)


src/components/NodeSelectorDialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const [initialRpc, setInitialRpc] = useState(rpc);
+  const setSavedRpc = useSetAtom(savedRpcNodeAtom);
+  const setSavedApi = useSetAtom(savedApiNodeAtom);
+
+  const save = () => {
+    assert(api && rpc);
+    setSavedApi(api);
+    setSavedRpc(rpc);
+    setIsOpen(false);
+    window.location.reload();
```
0xpatrickdev, line 53: Is a full page reload necessary here? I guess it doesn't matter much, as a majority of app state is dependent on the rpc + api endpoints.

0xpatrickdev (approved):
> It doesn't provide an entrypoint to the node selector outside of the error dialog, that can be added in a future PR. 

It would be nice to see this in the MVP, maybe as the last option in Network Dropdown and/or a url query param (`?showConnectionSettings=true`) for advanced users. But agree this can come later

src/components/NodeSelectorDialog.tsx
samsiegart: Yea it's just taking the easy way out. Otherwise we'd have to recreate the chainStorageWatcher and walletConnection, reset the state, and make sure we clean up all casting followers etc.

Issue #214, opened 2023-11-30
Update dapp-inter smart wallet provisioning UX to accommodate auto-wallet provisioning change


otoole-brendan:
## What is the Problem Being Solved?

In preparation for auto wallet provisioning where absent (https://github.com/Agoric/agoric-sdk/issues/8558) - there are dapp-inter changes we can make to accommodate this new flow/UX.

In current flow, a user is prompted with the 'You need to provision a wallet dialog that will cost 10BLD' if they connect Keplr on the app and no wallet is detected.

## Description of the Design

Once #8558 is deployed - the desired flow will change to:

1. User visits app.inter.trade
2. User connects Keplr
3. User configures new vault
4. User clicks 'Create Vault'
5. User is prompted with dialog that they will need 1IST in order to provision the wallet before being able to create the vault. They can opt to 'Proceed' (which will attempt to send the txn and should succeed if they have the IST required to provision and fail if they do not (Question @samsiegart can we customize the 'you don't have enough IST to provision a wallet' error message?)) , 'Get IST' or 'Cancel'. 'Get IST' brings up the Leap Elements modal to enable them to easily onboard IST. Hitting Cancel returns them to vault setup view.

Dialog Text + Design:

Title: 'Smart Wallet Required'
'To interact with contracts on the Agoric chain, a smart wallet must be created for your account. As an anti-spam measure, you will need 1 IST to fund its provision which will be deposited into the community fund. Click 'Proceed' to provision wallet and submit transaction.'

## Security Considerations
N/A
## Test Plan
Will be smoke tested
```[tasklist]
### Tasks
```

mhofman:
> In preparation for auto wallet provisioning where absent ([Agoric/agoric-sdk#8558](https://github.com/Agoric/agoric-sdk/issues/8558)) - dapp-inter changes are required in order to accommodate this new flow/UX.

To be clear, the changes are not needed before auto-provisioning is deployed. The current manual provisioning should still work. It actually is likely simpler to only deploy the UI change after auto-provision is deployed.

otoole-brendan:
Yep - apologies, I was aware of that having discussed with Sam yesterday. 'Required' is too strong...will update

PR #219, opened 2024-01-02
feat: auto wallet provision

samsiegart:
refs https://github.com/Agoric/dapp-inter/issues/214

## Screenshot
 
Screenshot 2024-01-02 at 12 59 57 AM


src/components/ProvisionSmartWalletNoticeDialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const smartWalletFeeForDisplay = smartWalletFee
+    ? smartWalletFee + ' IST'
+    : null;
+
+  const body = (
+    <span>
+      To interact with contracts on the Agoric chain, a smart wallet must be
+      created for your account. As an anti-spam measure, you will need{' '}
+      {smartWalletFeeForDisplay && <b>{smartWalletFeeForDisplay}</b>} to fund
+      its provision which will be deposited into the community fund. Click
```
mhofman, line 61: I'm rusty on this, is the "community fund" the same as the "reserve pool"?

I don't remember where explicit provision fees ended up in (I think also in the reserve pool). The new auto-provision mechanism charges like other swingset fees, and it ends up in the reserve pool.

mhofman (commented):
FYI, while 1 IST is charged for provisioning, 0.25 IST is still transferred from the provision pool to the user's account during provisioning, which means that effectively, only 0.75 IST will be deducted (but 1 IST is still needed). I don't think we want to go into that much details however.

otoole-brendan:
@samsiegart per recent Slack conversation - let's remove reference to community fund and change to

'....which will be deposited into the reserve pool.'

samsiegart:
> @samsiegart per recent Slack conversation - let's remove reference to community fund and change to
> 
> '....which will be deposited into the reserve pool.'

Fixed

0xpatrickdev (approved):

PR #220, opened 2024-01-08
Node selector button

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/211

Also improved a11y of node selector dialog by making it work with keyboard users. Before, if you wanted to add a custom node, you had to type it then click out of the input box to select it. If you pressed tab or enter, it would revert the custom input.

## Screenshots

image

image

### DIFF
--- a/src/components/Combobox.tsx
+++ b/src/components/Combobox.tsx
@@ -1,5 +1,5 @@
 import { Combobox as HeadlessComboBox, Transition } from '@headlessui/react';
-import { Fragment } from 'react';
+import { Fragment, useCallback, useState } from 'react';
 import { FiChevronDown } from 'react-icons/fi';
 
 const Combobox = ({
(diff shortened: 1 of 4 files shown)


src/components/Combobox.tsx
```diff
@@ -36,15 +57,23 @@ const Combobox = ({
...
                 <HeadlessComboBox.Option
                   key={option}
                   value={option}
-                  className="relative cursor-pointer select-none py-2 pl-4 pr-4 text-gray-900 hover:bg-gray-100"
+                  className={optionClasses}
                 >
                   {option}
                 </HeadlessComboBox.Option>
               ))}
+              {query && !options?.includes(query) && (
```
turadg, line 69: if `v` is falsy then `!v?.foo()` will be true, so this expression renders when `options` is falsy. I don't have a suggestion because I'm not clear on what is intended but I'll ask that the the logic be more explicit and clear, whatever is intended. Consider binding a const.

actually… why is `options` allowed to be undefined? what's a Combobox without any options? if you just change it to `options: string[]` then several lines in the component will be less shaky. When I tried this locally nothing was relying on the optionality.

src/components/Combobox.tsx
```diff
@@ -36,15 +57,23 @@ const Combobox = ({
             leaveTo="opacity-0"
           >
             <HeadlessComboBox.Options className="z-50 absolute mt-1 max-h-60 w-full overflow-auto rounded-md bg-white py-1 text-base shadow-lg ring-1 ring-black/5 focus:outline-none sm:text-sm">
-              {options?.map(option => (
+              {filteredOptions?.map(option => (
                 <HeadlessComboBox.Option
                   key={option}
                   value={option}
-                  className="relative cursor-pointer select-none py-2 pl-4 pr-4 text-gray-900 hover:bg-gray-100"
+                  className={optionClasses}
```
turadg, line 64: what magic is this⁉️ :) 

I'm looking in https://headlessui.com/react/combobox and don't see an API that takes a function for className. Are we on a different version?

mind making it work more transparently?

src/components/Combobox.tsx
samsiegart: Yea good point, just made options required.

The expression checks if a query exists, and if the query is not already contained in the list of options, then gives the user the option to "add" it if so.

src/components/Combobox.tsx
samsiegart: If you look at the interactive demo at the top and view "code" they use it there, it's also properly typed so I think it's transparent enough

Screenshot 2024-01-08 at 10 26 00 AM

turadg: The code I can find on that page uses a different function pattern,
```jsx
          <Combobox.Option key={person.id} value={person} as={Fragment}>
            {({ active, selected }) => (
              <li
                className={`${
                  active ? 'bg-blue-500 text-white' : 'bg-white text-black'
                }`}
```

but I am persuaded by the static typing. Thanks for the screenshot

turadg (approved):

PR #227, opened 2024-01-23
feat: deposit ist from provision dialog

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/244

Screenshot 2024-03-14 at 10 27 40 AM

### DIFF
--- a/src/components/ProvisionSmartWalletNoticeDialog.tsx
+++ b/src/components/ProvisionSmartWalletNoticeDialog.tsx
@@ -70,7 +70,7 @@ const ProvisionSmartWalletNoticeDialog = ({
     <>
       <div>
         To interact with contracts on the Agoric chain, a smart wallet must be
-        created for your account. As an anti-spam measure, you will need{' '}
+        created for your account. You will need{' '}
         {smartWalletFeeForDisplay && <b>{smartWalletFeeForDisplay}</b>} to fund
         its provision which will be deposited into the reserve pool. Click
         &quot;Proceed&quot; to provision wallet and submit transaction.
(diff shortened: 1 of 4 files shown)


otoole-brendan:
@samsiegart the one change I would make (and you might already be doing it but not visible here) is some logic around when a user can click 'Proceed'. I think it's better UX to not let them attempt the transaction by hitting 'Proceed' until they have at least 1IST balance to cover the smart wallet provisioning. So we should only illuminate the button when they have that required balance.

src/components/ProvisionSmartWalletNoticeDialog.tsx
```diff
@@ -55,18 +60,57 @@ const ProvisionSmartWalletNoticeDialog = ({
...
-      its provision which will be deposited into the reserve pool. Click
-      &quot;Proceed&quot; to provision wallet and submit transaction.
-    </span>
+    <>
+      <div>
+        To interact with contracts on the Agoric chain, a smart wallet must be
+        created for your account. As an anti-spam measure, you will need{' '}
+        {smartWalletFeeForDisplay && <b>{smartWalletFeeForDisplay}</b>} to fund
+        its provision which will be deposited into the reserve pool. Click
+        &quot;Proceed&quot; to provision wallet and submit transaction.
```
raphdev, line 80:
```suggestion
        To interact with contracts on the Agoric chain, a smart wallet must be
        created for your account. You will need{' '}
        {smartWalletFeeForDisplay && <b>{smartWalletFeeForDisplay}</b>} to fund
        its provision which will be deposited into the reserve pool. Click
        &quot;Proceed&quot; to provision wallet and submit transaction.
```

0xpatrickdev (approved):
Ran into #252 while testing, but appears to be unrelated. LGTM after the copy edit.

samsiegart:
@ivanlei I updated the text in the dialog per @raphdev 's suggestion, lmk if this looks good to you.

samsiegart:
> Ran into #252 while testing, but appears to be unrelated. LGTM after the copy edit.

The breaking change has now been [reverted](https://github.com/Agoric/dapp-inter/pull/253) and this PR has been rebased

PR #240, opened 2024-02-16
use Telescope-based cosmic-proto

turadg:
0.4.0 builds with Telescope, which provides some client support we were hand rolling


yarn.lock
```diff
@@ -1173,6 +1216,18 @@
     cosmjs-types "^0.8.0"
     long "^4.0.0"
 
+"@cosmjs/proto-signing@^0.32.2":
```
0xpatrickdev, line 1219: nit: it seems like we now have `0.30.x` and `0.32.x` versions for many of the `@cosmjs/*` deps

0xpatrickdev (approved):

yarn.lock
turadg: Good to raise. It looks to be mostly due to leapwallet pinning versions. We could use a resolution to override but that can be brittle.

I'll wait for @samsiegart 's input before merging. Also WRT test plan

src/vite-env.d.ts
```diff
@@ -45,8 +45,3 @@ declare module '@agoric/inter-protocol/src/vaultFactory/math' {
 declare module 'react-view-slider' {
   export const ViewSlider;
 }
-
-declare module '@agoric/cosmic-proto/swingset/query.js' {
```
samsiegart, line 49: Woohoo!

src/components/ProvisionSmartWalletNoticeDialog.tsx
```diff
@@ -12,14 +12,19 @@ const useSmartWalletFeeQuery = (rpc: string | null) => {
     const fetchParams = async () => {
       assert(rpc);
       try {
-        const params = await querySwingsetParams(rpc);
+        const client = await agoric.ClientFactory.createRPCQueryClient({
```
samsiegart, line 15: Remark: the wallet connection also makes its own query client, ideally at some point we could share between these two places somehow. I think the only main downside currently is an extra "status" query when you create an extra client.

samsiegart (approved):
LGTM, code is much cleaner and thanks for the descriptive commit messages

yarn.lock
samsiegart: I tested the deployed preview and looks like it worked. I don't know much about the version thing, it's good to be aware of though, maybe upgrading leap elements would fix it.

PR #243, opened 2024-02-23
fix: run `yarn upgrade ses`

LuqiPan:
Things I did to verify this:

`yarn dev` and navigate to http://localhost:5173 to verify that the UI can load now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "react-router-dom": "^6.4.5",
     "react-toastify": "^9.1.1",
     "react-view-slider": "^4.5.0",
-    "ses": "^1.0.1",
+    "ses": "^1.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,9 +14,9 @@
     "postinstall": "patch-package"
   },
   "dependencies": {
+    "@agoric/casting": "^0.4.3-u13.0",
```
samsiegart, line 17: I figure we should also change "ses" below from `"^1.0.1"` to `"^1.3.0"`

samsiegart (changes requested):

package.json
LuqiPan: Good call, updated

samsiegart (approved):

PR #251, opened 2024-03-12
feat: test cases for vaults ui

rabi-siddique:
The PR adds end-to-end test cases for:

- Connecting with Keplr Wallet
- Creating a new Vault
- Updating a Vault with different flows. 
- Closing a Vault. 

> ### How to execute tests:
> * Start the agoric chain with the following command:
> 
> ```
> docker run -d -p 26657:26657 -p 1317:1317 -p 9090:9090 ghcr.io/agoric/agoric-3-proposals:main
> ```
> 
> * Start the local server for `inter-dapp`
> 
> ```
> yarn dev
> ```
> 
> * Execute the tests using the following command:
> 
> ```
> yarn test:e2e
> ```


.gitignore
```diff
@@ -23,3 +23,8 @@ coverage
 *.njsproj
 *.sln
 *.sw?
+
+# Synpress directories and files
+.env
+tests
```
turadg, line 29: keeping `tests` out of SCM seems like a recipe for confusion. What is making those files? Can it be put somewhere else?

test/e2e/README.md
```diff
@@ -0,0 +1,55 @@
+# Running End-to-End Tests
+
+To run end-to-end test cases on your system, follow these steps:
```
turadg, line 3: why is `dapp-inter` documenting this?  shouldn't all this be in `@agoric/synpress` ?

.gitignore
rabi-siddique: @turadg The end-to-end test cases reside in the `test` folder. 

When we run these tests, a folder is automatically generated to store screenshots and videos of the test execution. Specifically, it creates two directories:

```
Screenshots folder: tests/e2e/screenshots
Videos folder: tests/e2e/videos
```

`tests` folder is excluded from SCM to avoid cluttering the repo with these large/temp files.

turadg: I get that. But it can be put somewhere else and I think that it should be to avoid confusion. It's quite surprising that `test` is in SCM and `tests` is ignored.

https://docs.cypress.io/guides/references/configuration#Screenshots says the default directory is `cypress/screenshots`. Why is this different?

test/e2e/README.md
rabi-siddique: It's more so of a quick guide to setting it up and running tests on `dapp-inter`. 
`@agoric/synpress` has more detailed documentation.

.gitignore
rabi-siddique: @turadg In `@agoric/synpress` repo, we follow a different folder structure for organizing our end-to-end test files. Specifically:
```
Screenshots folder: tests/e2e/screenshots
Videos folder: tests/e2e/videos
Tests folder: tests/e2e/spec
```
This structure follows the one used in the `synpress` repo, which we aimed to remain consistent with.
`inter-dapp` repo already has `test` folder. To have `tests` and `test` folder in the same repo, would have been confusing, so I went with adding end to end tests in `test/e2e/spec` directory.

test/e2e/README.md
turadg: > It's more so of a quick guide to setting it up and running tests on dapp-inter.

Then they shouldn't have to **Install Dependencies**, right?

And why should they have to paste some Docker command? I'd expect any commands they need to run for tests to begin with `yarn`.

.gitignore
turadg: i don't understand why Synpress changed the default paths that Cypress used, but I won't argue with it.

I do request that you ignore the directories that have large files, rather than everything that might appear in a `tests` directory. The confusion of gitignoring `tests` is already evident by virtue of this discussion.

```suggestion
screenshots
videos
```

rabi-siddique: @turadg Updated the `.gitignore`

.gitignore
```diff
@@ -23,3 +23,9 @@ coverage
 *.njsproj
 *.sln
 *.sw?
+
+# Synpress directories and files
+.env
+tests/e2e/screenshots
+tests/e2e/videos
```
turadg, line 30: why not all screenshots and videos? what if someone moves those directories, they'll have to know to update this.

rabi-siddique: @turadg Updated

turadg: That's still not all screenshots and videos. I don't understand the expectation that some screenshots or videos would ever be committed to this repo but given that I've asked three times not to have `tests` in gitignore, I assume you have a strong opinion about retaining it.

turadg (approved):

samsiegart (changes requested):
Looks like lint is failing, maybe adding an eslintrc like https://github.com/frazarshad/dapp-psm/blob/79bb3102a476b00ca685b75087dd54b659426e60/tests/e2e/.eslintrc.cjs should do the trick?

rabi-siddique:
> Looks like lint is failing, maybe adding an eslintrc like https://github.com/frazarshad/dapp-psm/blob/79bb3102a476b00ca685b75087dd54b659426e60/tests/e2e/.eslintrc.cjs should do the trick?

@samsiegart Thanks for pointing it out. I've addressed the linting issue. I had to add some eslint-specific dependencies and make code changes around that. Turns out, in `dapp-psm`, there weren't many linting issues because linter is applied to the `src` folder only and as a result, it never gets applied to the `tests` folder there. In `dapp-inter`, our linter isn't restricted to the `src` folder.

samsiegart:
We're close now, lint is passing but `yarn test` is failing:
```
⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/e2e/specs/test.spec.js [ test/e2e/specs/test.spec.js ]
ReferenceError: describe is not defined
 ❯ test/e2e/specs/test.spec.js:2:1
      1| /* eslint-disable ui-testing/no-disabled-tests */
      2| describe('Vaults UI Test Cases', () => {
       | ^
      3|   context('Test commands', () => {
      4|     let collateralizationRatio;

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

 Test Files  1 failed | 1 passed (2)
      Tests  16 passed (16)
```
I think we can solve this by ignoring the e2e tests in `vitest.config.ts` like: https://github.com/Agoric/dapp-psm/pull/125/commits/fff09f5e6106907a06f8d357e666217776997199

rabi-siddique:
> We're close now, lint is passing but `yarn test` is failing:
> 
> ```
> ⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯
> 
>  FAIL  test/e2e/specs/test.spec.js [ test/e2e/specs/test.spec.js ]
> ReferenceError: describe is not defined
>  ❯ test/e2e/specs/test.spec.js:2:1
>       1| /* eslint-disable ui-testing/no-disabled-tests */
>       2| describe('Vaults UI Test Cases', () => {
>        | ^
>       3|   context('Test commands', () => {
>       4|     let collateralizationRatio;
> 
> ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
> 
>  Test Files  1 failed | 1 passed (2)
>       Tests  16 passed (16)
> ```
> 
> I think we can solve this by ignoring the e2e tests in `vitest.config.ts` like: [Agoric/dapp-psm@fff09f5](https://github.com/Agoric/dapp-psm/commit/fff09f5e6106907a06f8d357e666217776997199)

@samsiegart Addressed. Thanks for the assistance. :)

samsiegart (approved):

PR #256, opened 2024-03-19
chore: suggest connection with chain before connecting with wallet

rabi-siddique:
Issue:
I've encountered an error after installing a new Keplr extension and visiting the dapp-inter URL. When attempting to connect with the wallet, I receive the following error:
image

This happens because when I imported or created my wallet, I didn't select the Agoric Chain. Also, there weren't any options available to select any testing chains:
A

I noticed that new users may have trouble connecting to the wallet. Typically, we can connect without any problems, but it seems that if we're already connected to the testing chains, it's easier. For example, when I was using the `offerup` dapp initially and then switched to `dapp-inter`, connecting my wallet was smooth because I was already connected to the Agoric chain during my work with `offerup` dapp.

What I've observed is that new users might face difficulty in connecting with the wallet and get the error message I shared above. 

So, to fix this, I've utilized the `suggestChain` utility from `@agoric/web-components`. Now, whenever we switch between testing networks, it prompts us to connect with the chain first if we're not already connected. I implemented this for the offerup dapp.

NOTE: This issue only occurs in our local environment; the production version works perfectly fine.

### DIFF
--- a/src/components/ConnectWalletButton.tsx
+++ b/src/components/ConnectWalletButton.tsx
@@ -10,6 +10,7 @@ import {
   networkConfigAtom,
   walletServiceAtom,
 } from 'store/app';
+import { suggestChain } from '@agoric/web-components';
 import clsx from 'clsx';
 import { useEffect } from 'react';
 import { useStore } from 'zustand';
@@ -53,6 +54,11 @@ const ConnectWalletButton = () => {
     return 'Connect Wallet';
   })();
 
+  const handleSuggestChainAndConnectWallet = async () => {
+    await suggestChain(url);
+    walletService.connect();
+  };
+
   const isDisabled = !chainStorageWatcher;
 
   return (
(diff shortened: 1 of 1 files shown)


src/components/ConnectWalletButton.tsx
```diff
@@ -53,6 +54,11 @@ const ConnectWalletButton = () => {
     return 'Connect Wallet';
   })();
 
+  const handleSuggestChainAndConnectWallet = async () => {
+    await suggestChain(url);
```
samsiegart, line 58: It might be better to have error handling, like showing a [toast](https://github.com/Agoric/dapp-inter/blob/main/src/service/wallet.ts#L141-L144), but given this only impacts testnets I don't think it's worth bothering.

samsiegart (approved):
Thanks for adding this! And yea as some context, the wallet-app does the suggest-chain thing, but psm and dapp-inter did not. So if you connected to emerynet on wallet-app before, you wouldn't notice this issue.  Doing suggest-chain from dapp-inter will make this process less tricky. 

Just tested locally against mainnet also and verified that it didn't break.

Issue #248, opened 2024-03-07
Minting Fee does not shown/applied in Vault Summary


otoole-brendan:
## What is the Problem Being Solved?

When a user creates a vault or adjusts a vault - a minting fee applies to the minted IST amount - which is added to the user's IST overall debt amount. This actual amount is not reflected in the 'Vault Summary' preview and so it's not clear to the user what amount that is. 

The dollar amount should be included as a line item so users' can clearly see the minting fee that is being incurred and added to their overall IST debt with the mint action they are taking

## Description of the Design

- [ ] For new vault creation -> Show calculated minting fee based on user-specific IST mint amount in the vault summary tile e.g. if minting fee 0.05% user opening 1000IST vault incurs 50IST charge which should be shown as ->  Minting Fee......0.5% (50 IST)
- [ ] For vault adjustment (+minting more IST action) -> add 'Minting Fee' and show adjustment as a line item between 'Debt' and 'Collateralization Ratio' e.g. if user is minting an additional 1000IST at 0.5% then show as  'Minting Fee (0.05%)................50IST (added to debt)' (@samsiegart this might be a bit clunky so let's discuss if there's a better way to illustrate this line item)

samsiegart:
In "Adjust Vault", the new debt total already includes the minting fee (see below, minting 100 IST):

<img width="444" alt="Screenshot 2024-03-07 at 2 15 15 PM" src="https://github.com/Agoric/dapp-inter/assets/8848650/6a9d258b-d9c5-4e6e-9540-81ad4f874533">

Maybe it's best to have an `(i)` tooltip explaining it's the total of `existing debt + newly minted + minting fee`. 

In "Create Vault", maybe we could show a line below minting fee, like "Debt After Minting Fee" or something:

<img width="438" alt="Screenshot 2024-03-07 at 2 18 58 PM" src="https://github.com/Agoric/dapp-inter/assets/8848650/35828d39-e75c-42ab-8598-23aee3ecb43e">

That would make it clearer that it's not charging you IST on the transaction, but rather it's being added to the debt.

otoole-brendan:
I knew the minting fee was being reflected to the new debt amount - it's more about better communicating to users a) that is the case but more importantly b) what that amount is. 

For create vault - yep I think I largely agree but call it 'Debt' or 'Your Debt Balance' and include tool tip explainer:

Depositing........................X stTIA
Minting.............................1000IST
[...]

Issue #247, opened 2024-03-07
Minting Fee text is confusing on adjust vault view (when minting more IST)


otoole-brendan:
## Describe the bug

When a user attempts to adjust a vault and mint more IST, they are served with the below text that is confusing since it refers to vault creation (their vault is already created).

'A minting fee of X will be charged upon vault creation'

## To Reproduce

1. Click on vault
2. Select 'Mint More' action from drop down
3. See text appear

## Expected behavior

Text should not reference vault creation, it should read:

'A <insert minting fee as specified by governance> minting fee will be charged on the newly minted IST amount and added to your overall IST debt.'

## Screenshots

<img width="694" alt="image" src="https://github.com/Agoric/dapp-inter/assets/93355546/77543b67-450f-49cf-ab24-fde8191fd272">

PR #257, opened 2024-03-20
Minting fee

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/247
fixes https://github.com/Agoric/dapp-inter/issues/248

### Adjust Vault Minting Fee Text
image

### Adjust Vault Tooltip + Minting Fee
image

### Create Vault Tooltip
create-vault-tooltip

### DIFF
--- a/src/components/AdjustVaultForm.tsx
+++ b/src/components/AdjustVaultForm.tsx
@@ -157,7 +157,7 @@ const AdjustVaultForm = () => {
                   vaultToAdjust?.params.mintFee,
                   2,
                 )}%
-          will be charged upon vault creation.`}
+                will be charged on the newly minted IST amount and added to your overall IST debt.`}
             </p>
           </div>
           <div className="col-span-2 lg:col-span-1">
--- a/src/components/AdjustVaultSummary.tsx
+++ b/src/components/AdjustVaultSummary.tsx
@@ -2,6 +2,7 @@ import clsx from 'clsx';
 import { useAtomValue } from 'jotai';
 import { displayFunctionsAtom } from 'store/app';
 import {
+  DebtAction,
   adjustVaultErrorsAtom,
   collateralActionAtom,
   collateralInputAmountAtom,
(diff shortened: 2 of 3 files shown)


samsiegart:
I made some adjustments apart from the latest comment on https://github.com/Agoric/dapp-inter/issues/248.

In particular, in "Adjust Vault", I didn't add the "Minting" line because there's nothing to show on the left of the arrow so it looked odd (we don't know the total amount minted from a particular vault, only its debt). Also, the contents of the tooltip as suggested: `Existing debt (1100) + minting fee (50) + accrued stability fee (X)` didn't seem quite right, because existing debt includes stability fee already, and we need to add the newly minted amount to that equation instead. Also, I didn't include the values because it seemed kinda noisy for a tooltip, but could add them back if needed.

otoole-brendan:
@samsiegart 

Some changes to improve consistency:

Adjust Vault: Change 'Debt' -> 'Your Debt Balance' (tool tip and its contents looks good so let's keep them)

Create Vault: Similarly the tool tip and contents work well here but I noticed the line items for create vault could do with re-ordering to more intuitive and to match 'Adjust Vault'. This includes:

- Move 'Collateralization Ratio' to be above the line
- Move 'Stability Fee' and 'Minimum Collateralization Ratio' below the line.

The rationale with the delineating line was to visually separate the static fees/rates (below the line) and the dynamic/vault-specific information (above the line). This might be a little out of the scope of this change but I think it's easier to address with this PR and does simplify UX nicely

samsiegart:
@otoole-brendan Thanks for taking a look, makes sense. Just adjusted as shown:

image

image

0xpatrickdev (approved):

otoole-brendan:
@samsiegart looks good but the order of line items below line for adjust vault and create vault differs. The ordering for adjust vault is best imo so can you change the create vault order to match adjust vault ordering? (just the below the line pieces)

samsiegart:
Updated the ordering in create-vault:

image

PR #261, opened 2024-04-01
ci: add daliy scheduled run for emerynet testing

frazarshad:
This PR does the following:
- It adjusts the previously written tests to work for both local network and emerynet. these updates include
    - changing the order of test to make the transactions cyclic. i.e. the amount put into the vault is returned in the end
    - update the wallet creation flow to use previously created wallet on local but to create a brand new wallet on emerynet
- it adds a scheduled run of the e2e_tests github action to run once everyday on emerynet 

**Note:** the behaviour of provisioning ATOM and IST from https://emerynet.faucet.agoric.net seems to be unreliable and flaky, but at the moment I am not aware of any other way to get these tokens in a new account

### DIFF
--- a/.github/workflows/reusable-workflow.yml
+++ b/.github/workflows/reusable-workflow.yml
@@ -12,6 +12,9 @@ jobs:
   e2e:
     runs-on: ubuntu-latest
 
+    env:
+      IS_EMERYNET_TEST: ${{ github.event_name == 'schedule' || inputs.network == 'emerynet' }}
+
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -33,6 +36,10 @@ jobs:
       - name: Run e2e tests
         run: ${{ inputs.docker_compose_command }}
         env:
+          # conditionals based on github event
+          SYNPRESS_PROFILE: ${{ env.IS_EMERYNET_TEST == 'true' && 'daily-tests' || 'synpress' }}
+          CYPRESS_AGORIC_NET: ${{ env.IS_EMERYNET_TEST == 'true' && 'emerynet' || 'local' }}
+          # for docker-compose.yml
           COMPOSE_DOCKER_CLI_BUILD: 1
           DOCKER_BUILDKIT: 1
           DOCKER_DEFAULT_PLATFORM: linux/amd64
--- a/test/e2e/specs/test.spec.js
+++ b/test/e2e/specs/test.spec.js
@@ -1,21 +1,15 @@
-import { mnemonics } from '../test.utils';
+import { mnemonics, phrasesList, MINUTE_MS } from '../test.utils';
+
 describe('Vaults UI Test Cases', () => {
   context('Test commands', () => {
-    const phrasesList = {
-      emerynet: {
-        interNetwork: 'Agoric Emerynet',
-        isLocal: false,
-      },
-      local: {
-        interNetwork: 'Local Network',
-        isLocal: true,
-      },
-    };
     const networkPhrases = phrasesList[Cypress.env('AGORIC_NET') || 'local'];
 
     it('should setup the wallet', () => {
       if (networkPhrases.isLocal) {
-        cy.setupWallet();
+        cy.setupWallet({
+          secretWords: mnemonics.user1,
+          walletName: 'user1',
+        });
       } else {
         const walletAddress = {
           value: null,
@@ -27,7 +21,7 @@ describe('Vaults UI Test Cases', () => {
         cy.origin('https://wallet.agoric.app/', () => {
           cy.visit('/wallet/');
 
-          cy.get('input.PrivateSwitchBase-input').click();
+          cy.get('input[type="checkbox"]').click();
           cy.contains('Proceed').click();
         });
         cy.acceptAccess();
@@ -60,6 +54,7 @@ describe('Vaults UI Test Cases', () => {
       }
     });
 
+    // eslint-disable-next-line ui-testing/missing-assertion-in-test
     it('should connect with the wallet', () => {
       cy.visit('/');
 
@@ -71,74 +66,12 @@ describe('Vaults UI Test Cases', () => {
       }
 
       cy.contains('Connect Wallet').click();
-      cy.acceptAccess().then(taskCompleted => {
-        expect(taskCompleted).to.be.true;
-      });
-      cy.get('label.cursor-pointer input[type="checkbox"]').check();
-      cy.contains('Proceed').click();
 
-      cy.acceptAccess().then(taskCompleted => {
-        expect(taskCompleted).to.be.true;
-      });
-    });
-
-    it('should create a new vault and approve the transaction successfully', () => {
-      cy.visit('/');
-      if (!networkPhrases.isLocal)
-        cy.get('button').contains('Keep using Old Version').click();
-
-      cy.contains('button', /ATOM/).click();
-
-      cy.contains('ATOM to lock up *')
-        .next()
-        .within(() => {
-          cy.get('input[type="number"]').click();
-          cy.get('input[type="number"]').clear();
-          cy.get('input[type="number"]').type(10);
-        });
-
-      cy.contains('button', 'Create Vault').click();
-
-      cy.confirmTransaction().then(taskCompleted => {
-        expect(taskCompleted).to.be.true;
-        cy.contains(
-          'p',
-          'You can manage your vaults from the "My Vaults" view.',
-        ).should('exist');
-      });
-      cy.get('label.cursor-pointer input[type="checkbox"]').check();
+      cy.acceptAccess();
+      cy.get('label input[type="checkbox"]').check();
       cy.contains('Proceed').click();
 
       cy.acceptAccess();
-    });
-
-    it('should adjust the collateral by performing a withdrawl and approve the transaction successfully', () => {
-      cy.contains('button', 'Manage my Vaults').click();
-      cy.contains('button', 'Back to vaults').click();
-      cy.contains('div', /ATOM.*#8/).click();
-
-        cy.contains('div', 'Adjust Collateral')
-          .next('.grid-cols-2')
-          .within(() => {
-            cy.contains('button', /^(Deposit|Withdraw|No Action)$/).click();
-            cy.contains('button', 'Withdraw').click();
-            cy.contains('.input-label', 'Amount')
-              .next('.input-wrapper')
-              .within(() => {
-                cy.get('input[type="number"]').click();
-                cy.get('input[type="number"]').type(5);
-              });
-          });
-
-        cy.contains('button', 'Adjust Vault').click();
-        cy.confirmTransaction().then(taskCompleted => {
-          expect(taskCompleted).to.be.true;
-          cy.contains('p', "Your vault's balances have been updated.", {
-            timeout: 60000,
-          }).should('exist');
-          cy.contains('Back to my vaults').click();
-        });
-      }
     });
 
     it('should create a new vault and approve the transaction successfully', () => {
@@ -148,14 +81,15 @@ describe('Vaults UI Test Cases', () => {
 
       if (networkPhrases.isLocal)
         cy.contains('button', 'Add new vault').click();
+
       cy.contains('button', /ATOM/).click();
 
-      cy.contains('.input-label', 'ATOM to lock up *')
+      cy.contains('ATOM to lock up *')
         .next()
         .within(() => {
           cy.get('input[type="number"]').click();
           cy.get('input[type="number"]').clear();
-          cy.get('input[type="number"]').type(2);
+          cy.get('input[type="number"]').type(10); // 2 in local
         });
 
       cy.contains('button', 'Create Vault').click();
@@ -165,7 +99,7 @@ describe('Vaults UI Test Cases', () => {
         cy.contains(
           'p',
           'You can manage your vaults from the "My Vaults" view.',
-          { timeout: 60000 },
+          { timeout: MINUTE_MS },
         ).should('exist');
         cy.contains('Manage my Vaults').click();
       });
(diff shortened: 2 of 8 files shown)


test/e2e/specs/test.spec.js
```diff
@@ -1,52 +1,177 @@
...
   context('Test commands', () => {
-    let collateralizationRatio;
-    it(`should setup Keplr account and connect with Agoric Chain`, () => {
-      cy.origin('https://wallet.agoric.app/', () => {
-        cy.visit('/');
-      });
-      cy.acceptAccess().then(taskCompleted => {
-        expect(taskCompleted).to.be.true;
-      });
+    const phrasesList = {
```
turadg, line 4: consider putting these into a config file

test/e2e/specs/test.spec.js
```diff
@@ -1,52 +1,177 @@
...
+        cy.setupWallet({
+          createNewWallet: true,
+          walletName: 'my created wallet',
+        });
+        cy.origin('https://wallet.agoric.app/', () => {
+          cy.visit('/wallet/');
 
-        cy.get('input.PrivateSwitchBase-input').click();
-        cy.contains('Proceed').click();
+          cy.get('input.PrivateSwitchBase-input').click();
```
turadg, line 30: this class name may change. please select by something stable in the UI. e.g. something the user sees

test/e2e/specs/test.spec.js
```diff
@@ -1,52 +1,177 @@
...
+
       cy.contains('Connect Wallet').click();
-      cy.get('label.cursor-pointer input[type="checkbox"]').check();
-      cy.contains('Proceed').click();
 
       cy.acceptAccess().then(taskCompleted => {
         expect(taskCompleted).to.be.true;
-        cy.acceptAccess().then(taskCompleted => {
+      });
+      cy.get('label.cursor-pointer input[type="checkbox"]').check();
```
turadg, line 77: needs a more stable selector

test/e2e/specs/test.spec.js
```diff
@@ -1,52 +1,177 @@
...
+    });
+
+    it('should open the new vault', () => {
+      cy.get('span')
+        .contains(/My Vaults.*\(\d+\)/)
+        .children()
+        .first()
+        .spread((...element) => {
+          // Get the total number of vaults present
+          const vaultCount = Number(element[0].innerHTML.slice(2, -1));
```
turadg, line 149: this looks fragile. maybe there's a regex that would work?

test/e2e/specs/test.spec.js
```diff
@@ -1,52 +1,177 @@
...
 
+          cy.get('div.shadow-card div.text-secondary:contains("#")')
```
turadg, line 151: fragile selector

test/e2e/specs/test.spec.js
```diff
@@ -1,52 +1,177 @@
...
+            .should('have.length', vaultCount)
+            .spread((...vaults) => {
+              expect(
+                vaults.filter(vaultNo => !/^#\d+$/.test(vaultNo.innerHTML)),
+              ).to.be.empty;
+              // Get the vault with the largest number and click on it
```
turadg, line 157: the largest number meaning the last one? is it also last in the list? that would be a more stable query and simpler implementation

test/e2e/specs/test.spec.js
```diff
@@ -56,16 +181,16 @@ describe('Vaults UI Test Cases', () => {
...
         });
 
       cy.contains('button', 'Adjust Vault').click();
       cy.confirmTransaction().then(taskCompleted => {
         expect(taskCompleted).to.be.true;
-        cy.contains('p', "Your vault's balances have been updated.").should(
-          'exist',
-        );
+        cy.contains('p', "Your vault's balances have been updated.", {
+          timeout: 60000,
```
turadg, line 192: why are these the timeouts? are they particular to each of these tests? consider putting in a config file. E.g. this one is `MINUTE_MS`

turadg (changes requested):

test/e2e/specs/test.spec.js
frazarshad: unfortunately no, these are sorted by the ID of the vault not by the vault #
there exists a case on local chain where the latest (and largest) vault is the third last item

test/e2e/specs/test.spec.js
frazarshad: these timeouts are specifically placed after a transactions occurs, since on average a transaction takes more than the default 30s but less than 60s

test/e2e/specs/test.spec.js
turadg: > the ID of the vault not by the vault #

By "vault #" you mean like `#2` here?

Screenshot 2024-04-02 at 6 58 24 AM

That is the unique ID of the vault in the namespace of its collateral type. The ID is a serial number so the highest ID is the most recent vault (of that collateral type).

> exists a case on local chain where the latest (and largest) vault is the third last item

Ok, so they're not sorted by increasing ID or size. I understand now that you want the latest vault, meaning the one with the largest #ID. Please clarify "largest number" because there are many numbers on a vault card.

test/e2e/specs/test.spec.js
```diff
@@ -1,52 +1,168 @@
...
+              // Get the vault with the largest number and click on it
+              const maxValue = vaults.reduce(
+                (maxValue, currentValueRaw) => {
+                  const currentValue = Number(
+                    currentValueRaw.innerHTML.slice(1),
+                  );
+                  return currentValue > maxValue ? currentValue : maxValue;
+                },
+                Number(vaults[0].innerHTML.slice(1)),
+              );
```
turadg, line 159: If I'm following your code right, this would be clearer:
```suggestion
              const vaultIds = vaults.map(v => Number(v.innerHTML.slice(1)));
              const latestId = Math.max(...vaultIds);
```

test/e2e/specs/config.js
```diff
@@ -0,0 +1,17 @@
...
+    interNetwork: 'Agoric Emerynet',
+    isLocal: false,
+  },
+  local: {
+    interNetwork: 'Local Network',
+    isLocal: true,
+  },
+};
+
+module.exports = {
```
turadg, line 14: do these have to be CJS modules? we've been using MJS whenever we can

test/e2e/specs/test.spec.js
```diff
@@ -1,52 +1,161 @@
...
+      cy.acceptAccess();
+    });
+
+    it('should withdraw ATOMS from vault successfully', () => {
+      if (networkPhrases.isLocal) {
+        cy.visit('/');
+        cy.contains('div', /ATOM.*#5/).click();
+
+        cy.contains('div', 'Adjust Collateral')
+          .next('.grid-cols-2')
```
turadg, line 81: still some implementation dependence but we can revisit if/when it breaks

turadg (approved):

frazarshad:
Currently due to the fact that the IST provisioning step for emerynet is flaky and prone to failure, I think we should avoid merging this until that issue is fixed.
Marking as draft to avoid accidental merges

.github/workflows/reusable-workflow.yml
```diff
@@ -33,6 +36,10 @@ jobs:
       - name: Run e2e tests
         run: ${{ inputs.docker_compose_command }}
         env:
+          # conditionals based on github event
+          SYNPRESS_PROFILE: ${{ env.IS_EMERYNET_TEST == 'true' && 'daily-tests' || 'synpress' }}
+          CYPRESS_AGORIC_NET: ${{ env.IS_EMERYNET_TEST == 'true' && 'emerynet' || 'local' }}
+          # for docker-compose.yml
```
rabi-siddique, line 42: I think `AGORIC_NET` is enough

test/e2e/specs/config.js
```diff
@@ -0,0 +1,12 @@
...
+
+export const phrasesList = {
+  emerynet: {
+    interNetwork: 'Agoric Emerynet',
+    isLocal: false,
+  },
+  local: {
+    interNetwork: 'Local Network',
+    isLocal: true,
+  },
```
rabi-siddique, line 11: There is a file called `test.utils.js`. Perhaps these declarations can be placed there?

test/e2e/specs/test.spec.js
```diff
@@ -1,33 +1,84 @@
...
-    it('should set up wallet', () => {
-      // Using exports from the synthetic-chain lib instead of hardcoding mnemonics UNTIL https://github.com/Agoric/agoric-3-proposals/issues/154
-      cy.setupWallet({
-        secretWords: mnemonics.user1,
-        walletName: 'user1',
-      }).then(taskCompleted => {
-        expect(taskCompleted).to.be.true;
-      });
+    const networkPhrases = phrasesList[Cypress.env('AGORIC_NET') || 'local'];
+
```
rabi-siddique, line 7: Just to confirm, are we referring to `CYPRESS_AGORIC_NET` over here?

.github/workflows/reusable-workflow.yml
frazarshad: the `CYPRESS_` prefix is required to be able to use it inside our cypress tests https://docs.cypress.io/guides/guides/environment-variables#Option-3-CYPRESS_

test/e2e/specs/test.spec.js
```diff
@@ -1,33 +1,83 @@
...
+        cy.origin('https://wallet.agoric.app/', () => {
+          cy.visit('/wallet/');
+
+          cy.contains(/agoric.{39}/).spread(element => {
+            return element.innerHTML.match(/agoric.{39}/)[0];
+          });
+        }).then(address => {
+          walletAddress.value = address;
+        });
+        cy.origin(
```
rabi-siddique, line 38: Are we trying to get wallet address in this snippet?
```
        cy.origin('https://wallet.agoric.app/', () => {
          cy.visit('/wallet/');

          cy.get('input[type="checkbox"]').click();
          cy.contains('Proceed').click();
        });
        cy.acceptAccess();

        cy.origin('https://wallet.agoric.app/', () => {
          cy.visit('/wallet/');

          cy.contains(/agoric.{39}/).spread(element => {
            return element.innerHTML.match(/agoric.{39}/)[0];
          });
        }).then(address => {
          walletAddress.value = address;
        });
        
 ```
 
 If yes, I believe we created a method in `@agoric/synpress` for the same purpose.

File: test/e2e/specs/test.spec.js [hunk 15]
```diff
@@ -1,33 +1,83 @@
...
+            cy.get('[type="submit"]').first().click();
+            cy.get('body').contains('success').should('exist');
+
+            cy.visit('/');
+            cy.get('[id="address"]').first().type(walletAddress.value);
+            cy.get('[type="radio"][value="client"]').click();
+            cy.get('[type="submit"]').first().click();
+            cy.get('body').contains('success').should('exist');
+          },
+        );
```
rabi-siddique, line 53: Do you think we should abstract this wallet setup for emerynet as a custom command in `support.js`? I feel that it would be easier to maintain.

test/e2e/specs/test.spec.js
```diff
@@ -1,33 +1,83 @@
...
     it('should connect with the wallet', () => {
       cy.visit('/');
 
+      if (!networkPhrases.isLocal) cy.contains('button', 'Dismiss').click();
+      cy.get('button').contains('Local Network').click();
+      cy.get('button').contains(networkPhrases.interNetwork).click();
+      if (!networkPhrases.isLocal)
+        cy.get('button').contains('Keep using Old Version').click();
+
       cy.contains('Connect Wallet').click();
```
rabi-siddique, line 67: Is 61 to 66 for connecting with emerynet? For local network connection, line 62 and 63 seem redundant? 
Do you think we should do it like:
```js
 if (!networkPhrases.isLocal) {
 cy.contains('button', 'Dismiss').click();
 cy.get('button').contains('Local Network').click();
 cy.get('button').contains(networkPhrases.interNetwork).click(); 
cy.get('button').contains('Keep using Old Version').click();
 }
```

I think by doing this it becomes distinguishable about what's different in local and emerynet

rabi-siddique: `if (!networkPhrases.isLocal) cy.contains('button', 'Dismiss').click();` Do we need this? Haven't encountered this

test/e2e/specs/test.spec.js
frazarshad: this is for getting the wallet address _before_ keplr has connected to anything. this means we cannot use the function in `@agoric/synpress` for this purpose

test/e2e/test.utils.js
```diff
@@ -10,3 +10,16 @@ export const accountAddresses = {
   gov1: 'agoric1ee9hr0jyrxhy999y755mp862ljgycmwyp4pl7q',
   gov2: 'agoric1wrfh296eu2z34p6pah7q04jjuyj3mxu9v98277',
 };
+
+export const MINUTE_MS = 60000;
+
```
rabi-siddique, line 15: `export const DEFAULT_TIMEOUT = 1 * 60 * 1000;` is more semantic. 
Given we're using `MINUTE_MS` at multiple places. Perhaps it should be called `DEFAULT_TIMEOUT`?

test/e2e/specs/test.spec.js
frazarshad: it only occurs when the dapp cant find a local chain, so we dismiss the "Error connecting" dialog box in emerynet tests.
You might not have encountered this if your local chain was running

test/e2e/specs/test.spec.js
```diff
@@ -45,15 +95,37 @@ describe('Vaults UI Test Cases', () => {
...
-      cy.contains('button', 'Manage my Vaults').click();
-      cy.contains('button', 'Back to vaults').click();
-      cy.contains('div', /ATOM.*#8/).click();
+    it('should open the new vault', () => {
+      cy.get('span')
+        .contains(/My Vaults.*\(\d+\)/)
+        .children()
+        .first()
+        .spread(element => {
+          // Get the total number of vaults present
```
rabi-siddique, line 110: Why use `spread`? `spread` is useful when we have an array of elements to process? And over here we only process the `.first()`? Do you think we should use `.then()`?

test/e2e/specs/test.spec.js
```diff
@@ -45,15 +95,37 @@ describe('Vaults UI Test Cases', () => {
...
+          const vaultCountRegex = element.innerHTML.match(/\((\d+)\)/);
+          const vaultCount = Number(vaultCountRegex[1]);
+
+          cy.findAllByText(/#\d+/)
+            .should('have.length', vaultCount)
+            .spread((...vaults) => {
+              expect(
```
rabi-siddique, line 117: I think we can use `.then` over here as well. `.then(vaults => {` would suffice

test/e2e/specs/test.spec.js
```diff
@@ -45,15 +95,37 @@ describe('Vaults UI Test Cases', () => {
...
+        cy.contains('Manage my Vaults').click();
       });
     });
 
-    it('should adjust the collateral by performing a withdrawl and approve the transaction successfully', () => {
-      cy.contains('button', 'Manage my Vaults').click();
-      cy.contains('button', 'Back to vaults').click();
-      cy.contains('div', /ATOM.*#8/).click();
+    it('should open the new vault', () => {
+      cy.get('span')
```
rabi-siddique, line 105: I think since we are testing for adjusting collateral, we can move the code in this test case to the subsequent test case. Or if you think we will reuse this code at places, perhaps we create a custom command for it?

test/e2e/specs/test.spec.js
frazarshad: it is readable enough as is. i dont think moving the function for a single use will improve clarity

test/e2e/test.utils.js
frazarshad: i think the name of this variable gives a clear enough indication of the number it represents. it can be left as is

test/e2e/specs/test.spec.js
frazarshad: we arent reusing the code at any other place and it is only being triggered once. can you explain further what you mean?

test/e2e/specs/test.spec.js
frazarshad: spread is required here because then only passes the first matching elements as an argument

rabi-siddique (approved):

test/e2e/specs/test.spec.js
rabi-siddique: I am suggesting that we should not have this as a separate test case. We are not testing `should open the new vault`. Instead we are testing `should adjust the collateral by performing a withdrawl and approve the transaction successfully`.

test/e2e/specs/test.spec.js
rabi-siddique: Won't use the same setup for liquidation test cases? Setting the wallet will be same with those test cases as well. Perhaps, we can reuse this snippet.

Issue #260, opened 2024-03-29
Add stkATOM icon and stkATOM default for leap elements


otoole-brendan:
## What is the Problem Being Solved?
As part of the collateral onboarding process for stkATOM - we need to add the stkATOM icon to dapp UI so it renders throughout.

## Description of the Design

1) Configure dapp-inter UI to render stkATOM icon from Cosmos chain registry https://github.com/cosmos/chain-registry/tree/master/persistence/images 

2) Add stkATOM default to Leap Elements

## Security Considerations
N/A
## Test Plan
Visually confirm in UI

PR #262, opened 2024-04-08
feat: add stkATOM

0xpatrickdev:
closes: #260


src/components/leap-elements/LiquidityModal.tsx
```diff
@@ -30,6 +30,8 @@ const chainIdForCollateralPetname = (petname?: string) => {
       return 'cosmoshub-4';
     case 'stTIA':
       return 'stride-1';
+    case 'stkATOM':
+      return 'core-1';
```
0xpatrickdev, line 34: https://github.com/cosmos/chain-registry/blob/5a2352d22de715e0115fc64887944993bea8e123/persistence/chain.json#L4

turadg (approved):
LGTM

src/components/leap-elements/LiquidityModal.tsx
turadg: thanks, I had to do some google searching without this

PR #266, opened 2024-04-17
test: add test for verifying https://app.inter.trade url

rabi-siddique:


samsiegart (commented):
I'm not sure that testing that the prod URL is live in CI is desirable, is there more context about the motivations for this? I could see a test in https://github.com/Agoric/dapp-inter-jumper being more useful for loading a local version of the UI and verifying that it redirects to the appropriate [ReferencedUI](https://p2p-org.github.io/p2p-agoric-vstorage-viewer/#https://main.rpc.agoric.net:443|published,published.vaultFactory|published.vaultFactory.governance)

rabi-siddique:
> I'm not sure that testing that the prod URL is live in CI is desirable, is there more context about the motivations for this? I could see a test in https://github.com/Agoric/dapp-inter-jumper being more useful for loading a local version of the UI and verifying that it redirects to the appropriate [ReferencedUI](https://p2p-org.github.io/p2p-agoric-vstorage-viewer/#https://main.rpc.agoric.net:443%7Cpublished,published.vaultFactory%7Cpublished.vaultFactory.governance)

@samsiegart should we make it part of the periodic run of dapp-inter test cases with emerynet over here #261

samsiegart:
> @samsiegart should we make it part of the periodic run of dapp-inter test cases with emerynet over here https://github.com/Agoric/dapp-inter/pull/261

I think it would be best to put these in dapp-inter-jumper and run as part of CI against a local build, and also against emerynet periodically. Otherwise, yea I'm okay with just running them periodically against emerynet/live URL. However, don't the test cases in https://github.com/Agoric/dapp-inter/pull/261 run in CI **as well as** periodically? Or can they be separated?

rabi-siddique:
> > @samsiegart should we make it part of the periodic run of dapp-inter test cases with emerynet over here #261
> 
> I think it would be best to put these in dapp-inter-jumper and run as part of CI against a local build, and also against emerynet periodically. Otherwise, yea I'm okay with just running them periodically against emerynet/live URL. However, don't the test cases in #261 run in CI **as well as** periodically? Or can they be separated?

Yes, test cases in #261 run in CI periodically. Separated as in?

rabi-siddique:
> > Separated as in?
> 
> It looks like we have one [workflow file](https://github.com/Agoric/dapp-inter/blob/472d22ea5730daf129e19648498a5eff26e55b92/.github/workflows/e2e_tests.yml) that runs both in PRs and periodically. I'm suggesting that for the test cases that depend on live versions of apps, rather than locally built versions, we create a separate workflow for them so that they don't run in PRs.

Makes sense. Agreed! I'll discuss with Fraz

rabi-siddique:
Turning this PR into a draft until #261 is merged. The test in this PR will be made part of the periodic CI run with emerynet.

PR #272, opened 2024-04-22
test: bash scripts for liquidation operations with vaults

rabi-siddique:
The PR adds bash scripts for creating vaults, changing prices, placing bids, and viewing auctions. 

These scripts will be used in a subsequent PR for liquidation testing. 

To test these scripts locally, make sure your local chain is up and running. Run the following command:
```
docker run -d -p 26657:26657 -p 1317:1317 -p 9090:9090 ghcr.io/agoric/agoric-3-proposals:latest

```

Make sure you are present in the root directory of `dapp-inter`. You can test the scripts randomly with different values. 

- Testing creating vaults. To create a vault, run the script:
```
~/dapp-inter/test/e2e/test-scripts/create-vaults.sh 400 70 gov1 
```
This will create a vault that submits `70 ATOMs` as collateral and mints `400 ISTs` We will later use these ISTs to place bids. 

- Testing changing price. For example:
```
~/dapp-inter/test/e2e/test-scripts/set-oracle-price.sh 9.99
```
In this script, the parameter 9.99 represents the desired price value you want to update in the oracle. 

- Testing placing bids. Run the `place-bids.sh` script to place bids from the `gov1` account:
```
~/dapp-inter/test/e2e/test-scripts/place-bids.sh
```

- Testing auction. You can test presence of different auction fields and their values following this example:
```
~/dapp-inter/test/e2e/test-scripts/view-auction.sh "book0.collateralAvailable" "0 ATOM"
```

In this example, `book0.collateralAvailable` is the field we are looking for and `0 ATOM` is the expected value.

### DIFF
--- a/test/e2e/test-scripts/common.sh
+++ b/test/e2e/test-scripts/common.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+# Common variables
+export mnemonicGov1="such field health riot cost kitten silly tube flash wrap festival portion imitate this make question host bitter puppy wait area glide soldier knee"
+export mnemonicGov2="physical immune cargo feel crawl style fox require inhale law local glory cheese bring swear royal spy buyer diesel field when task spin alley"
+export mnemonicUser1="tackle hen gap lady bike explain erode midnight marriage wide upset culture model select dial trial swim wood step scan intact what card symptom"
+export gov1AccountAddress=agoric1ee9hr0jyrxhy999y755mp862ljgycmwyp4pl7q
+export gov2AccountAddress=agoric1wrfh296eu2z34p6pah7q04jjuyj3mxu9v98277
+export user1Address=agoric1ydzxwh6f893jvpaslmaz6l8j2ulup9a7x8qvvq
+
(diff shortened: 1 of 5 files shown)


turadg (commented):
Bash scripts are harder to maintain than code modules. Given most Agoric code is Javascript, could these be too?

NodeJS has pretty good scripting support built in. You could also use something like `execa` if you want more shell-ish usage.

rabi-siddique:
> Bash scripts are harder to maintain than code modules. Given most Agoric code is Javascript, could these be too?
> 
> NodeJS has pretty good scripting support built in. You could also use something like `execa` if you want more shell-ish usage.

I'll explore it and let you know about it. `execa` seems cool. Will try to write a couple of test cases using `execa` and see if we can bring it in.

rabi-siddique:
I am no longer using bash scripts for liquidation tests. PR #275 introduces tests that do not rely on bash scripts. Therefore, I'm closing this pull request. 

@turadg Thanks for the review and helpful insight on using JS instead of bash scripts.

PR #274, opened 2024-04-24
ci: split docker compose and add reusable E2E workflow

rabi-siddique:
## Changes
- Split the docker-compose file into shared(`docker-compose-base.yml`) and specific(`docker-compose-vaults.yml`) configurations.
- Added a reusable workflow for end-to-end (E2E) tests.

## Motivation
While working on Liquidation tests in PR #268, I noticed that the CI process takes too long because we were running both Vaults and Liquidation test cases in a single GitHub Actions runner. To optimize our CI workflow, this PR enables parallel GitHub Action runners for running Vaults and Liquidation(when #268 is merged) E2E tests separately. This can save us time and simplify the management of our tests and CI workflows.

I've already implemented these changes in PR #268, but I want to keep that PR focused and prevent it from becoming too large. Once this PR is merged, I can rebase PR #268 accordingly.

### DIFF
--- a/test/e2e/docker-compose-base.yml
+++ b/test/e2e/docker-compose-base.yml
@@ -35,24 +35,11 @@ services:
       - GITHUB_BASE_REF=${GITHUB_BASE_REF}
       - GITHUB_HEAD_REF=${GITHUB_HEAD_REF}
       - SECRET_WORDS="orbit bench unit task food shock brand bracket domain regular warfare company announce wheel grape trust sphere boy doctor half guard ritual three ecology"
-    depends_on:
-      - display
-      - video
-      - agd
     entrypoint: []
     working_dir: /app
     volumes:
       - ./docker/videos:/app/test/e2e/videos
       - ./docker/screenshots:/app/test/e2e/screenshots
-    command: >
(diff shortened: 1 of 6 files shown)


test/e2e/docker-compose-base.yml
```diff
@@ -35,24 +35,11 @@ services:
       - GITHUB_BASE_REF=${GITHUB_BASE_REF}
       - GITHUB_HEAD_REF=${GITHUB_HEAD_REF}
       - SECRET_WORDS="orbit bench unit task food shock brand bracket domain regular warfare company announce wheel grape trust sphere boy doctor half guard ritual three ecology"
-    depends_on:
-      - display
-      - video
-      - agd
     entrypoint: []
```
rabi-siddique, line 38: I initially planned to include `depends_on` in `docker-compose-base.yml`, but discovered that it's not a shared configuration across services:
https://docs.docker.com/compose/multiple-compose-files/extends/#exceptions-and-limitations

turadg:
This doesn't affect development or maintenance of the dapp so @samsiegart and I don't have any special insight into it. I think @frazarshad would be a suitable approver.

frazarshad (approved):

PR #275, opened 2024-04-25
test: initial test cases for liquidation

rabi-siddique:
The PR introduces initial test cases for the happy path in the Liquidation workflow. It also includes the CI setup for running these tests in the CI pipeline.


package.json
```diff
@@ -85,6 +85,7 @@
   },
   "resolutions": {
     "**/@agoric/xsnap": "0.14.3-dev-9f085d3.0",
-    "**/@agoric/time": "0.3.3-dev-97cb715.0"
+    "**/@agoric/time": "0.3.3-dev-97cb715.0",
```
turadg, line 88: could these all be `ut15` now?

test/e2e/docker-compose-liquidation.yml
```diff
@@ -0,0 +1,54 @@
...
+      - display
+      - video
+      - agd
+    command: >
+      bash -c 'echo -n "======> local noVNC URL: http://localhost:8080/vnc.html?autoconnect=true " &&
+      yarn wait-on http://display:8080 &&
+      echo -n "======> remote noVNC URL: " &&
+      curl -s ngrok:4040/api/tunnels | jq -r .tunnels[0].public_url &&
+      nginx &&
+      npx start-server-and-test "yarn dev" http-get://localhost:5173 "yarn test:e2e --spec test/e2e/specs/liquidation.spec.js"'
```
turadg, line 17: `npx` will hit the external network to fetch. why not make it a devDependency?

test/support.js
```diff
@@ -1 +1,54 @@
...
+Cypress.Commands.add('setOraclePrice', price => {
+  cy.exec(
+    `agops oracle setPrice --keys gov1,gov2 --pair ATOM.USD --price ${price} --keyring-backend=test`,
+  ).then(({ stdout }) => {
+    expect(stdout).to.not.contain('Error');
+    expect(stdout).to.not.contain('error');
+  });
+});
+
+Cypress.Commands.add('createVault', params => {
```
turadg, line 37: these look good. I hope to eventually provide a JS API from the agoric-cli package or similar so this doesn't have to shell out to a JS program.

.eslintignore
```diff
@@ -3,3 +3,5 @@ coverage
 dist
 patches
 yarn.lock
+test
```
turadg, line 6: what are the lint failures? If you think some are wrong, instead of disabling lint we should tailor the rules to the ones that should be followed.

test/e2e/specs/liquidation.spec.js
```diff
@@ -54,6 +60,7 @@ describe('Wallet App Test Cases', () => {
       });
       cy.setOraclePrice(12.34);
     });
+    /* eslint-disable ui-testing/missing-assertion-in-test */
```
turadg, line 63: this is legit, right? perhaps you're testing that the function executes without throwing. does Cypress have something like Ava's `t.pass();` for if the line is reached?

test/e2e/support.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable ui-testing/no-disabled-tests */
```
turadg, line 1: I don't see any disabled tests. Is this rule buggy?

turadg (approved):

.eslintignore
rabi-siddique: Thanks for catching that. I had linting issues in `support.js` because it wasn't in the `test/e2e` directory, where the synpress-specific eslint file is located.

test/e2e/docker-compose-liquidation.yml
rabi-siddique: It is a devDependency:
https://github.com/rabi-siddique/rs-dapp-inter/blob/rs-initial-test-cases-for-liquidation/package.json#L78

test/e2e/support.js
rabi-siddique: Removed it

package.json
turadg: I meant 'u15' which is the version segment for upgrade-15 release. I see the new dep is using that.

test/e2e/docker-compose-liquidation.yml
turadg: Ah then no need for npx. 'npm run'

package.json
rabi-siddique: Ohh. Thanks for clarifying. :) Updated

test/e2e/specs/liquidation.spec.js
rabi-siddique: Didn't find anything like `t.pass()` :D

PR #276, opened 2024-04-25
test: add tests for placing bids and make vaults enter liquidation

rabi-siddique:


test/e2e/specs/liquidation.spec.js
```diff
@@ -84,4 +84,51 @@ describe('Wallet App Test Cases', () => {
...
+
+    it('should change ATOM price to 9.99', () => {
+      cy.setOraclePrice(9.99);
+    });
+
+    it('switch to user1 wallet', () => {
+      cy.switchWallet('user1');
+    });
+
+    it('should verify vaults that are at a risk of being liquidated', () => {
```
turadg, line 128: changing the wallet and the price aren't being tested, right?
```suggestion
    it('should verify vaults that are at a risk of being liquidated', () => {
      cy.switchWallet('user1');
      cy.setOraclePrice(9.99);
```

rabi-siddique: Makes sense. Changing it after testing it locally.

rabi-siddique: Updated

turadg (approved):

PR #281, opened 2024-05-17
refactor: simplify cypress command for adding keys from shell

rabi-siddique:
The PR does the following:
- Simplifies the cypress custom command for adding keys from the shell. 
- Remove installing `expect` package in the `Dockerfile`. This package was previously used with `agd keys add` command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ENV PATH="/usr/src/agoric-sdk/packages/agoric-cli/bin:${PATH}"
 
 # Install necessary dependencies
 RUN apt-get update \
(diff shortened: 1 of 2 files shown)


turadg (approved):
fwiw I'd call it a `refactor` rather than `chore`

rabi-siddique:
> fwiw I'd call it a `refactor` rather than `chore`

Makes sense. I'll reword it before merging. :)

rabi-siddique:
> > fwiw I'd call it a `refactor` rather than `chore`
> 
> Makes sense. I'll reword it before merging. :)

re: done

PR #282, opened 2024-05-17
test: liquidate vaults and viewing auction data

rabi-siddique:
The PR adds tests to ensure vaults are in a `liquidating` state and verifies the auction data. 

It also adds tests for verifying if a vault has reached a `liquidated` state but I'm currently deferring the execution of these tests in CI after updating the a3p Docker container, as I've noticed a discrepancy in behavior.

### DIFF
--- a/test/e2e/specs/liquidation.spec.js
+++ b/test/e2e/specs/liquidation.spec.js
@@ -1,4 +1,9 @@
-import { mnemonics, accountAddresses } from '../test.utils';
+import {
+  mnemonics,
+  accountAddresses,
+  LIQUIDATING_TIMEOUT,
+  LIQUIDATED_TIMEOUT,
+} from '../test.utils';
 
 describe('Wallet App Test Cases', () => {
   context('Setting up accounts', () => {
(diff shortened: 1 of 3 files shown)


test/e2e/.eslintrc.cjs
```diff
@@ -3,4 +3,7 @@ const synpressPath = path.join(process.cwd(), '/node_modules/@agoric/synpress');
 
 module.exports = {
   extends: `${synpressPath}/.eslintrc.js`,
+  rules: {
+    'cypress/unsafe-to-chain-command': 'off',
```
turadg, line 7: why is this necessary? what are consequences?

please justify in a comment

test/e2e/specs/liquidation.spec.js
```diff
@@ -124,5 +129,63 @@ describe('Wallet App Test Cases', () => {
...
+    it('should verify the value of collateralAvailable from the CLI successfully', () => {
+      const fieldName = 'book0.collateralAvailable';
+      const expectedValue = '45 ATOM';
+
+      cy.checkFieldValue(fieldName, expectedValue).then(isMatched => {
+        expect(isMatched).to.be.true;
+      });
+    });
+
+    it.skip('should wait and verify vaults are liquidated', () => {
```
turadg, line 175: please justify/explain the skipping in a comment

test/e2e/specs/liquidation.spec.js
```diff
@@ -124,5 +129,63 @@ describe('Wallet App Test Cases', () => {
...
+      });
+    });
+
+    it('should verify the value of startPrice from the CLI successfully', () => {
+      const fieldName = 'book0.startPrice';
+      const expectedValue = '9.99 IST/ATOM';
+
+      cy.checkFieldValue(fieldName, expectedValue).then(isMatched => {
+        expect(isMatched).to.be.true;
+      });
```
turadg, line 145: could the callback be an async function?
```suggestion
      expect(await cy.checkFieldValue(fieldName, expectedValue)).to.be.true;
```

or could `checkFieldValue` be an `assertFieldValue` and throw when they don't match?

rabi-siddique: Cypress tends to have trouble with async awaits. Please see [this](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress#The-Cypress-Command-Queue).

test/e2e/specs/liquidation.spec.js
rabi-siddique: The tests were running smoothly on my computer, but they failed in CI. I tried to replicate the CI environment on my local machine by updating a3p to latest version. Strangely, after updating, the tests failed on my local machine as well. I discovered that the tests pass when I use an older version of the container, specifically this one: `ghcr.io/agoric/agoric-3-proposals@sha256:c7cb71fb71227a52d407a2df7366571635c09a4acb4a30dc19a322ea9f9c6526`.

test/e2e/.eslintrc.cjs
rabi-siddique: If I keep this rule on, I get this error with some commands that are chained:

```
error    It is unsafe to chain further commands that rely on the subject after this command. It is best to split the chain, chaining again from `cy.` in a next command line  cypress/unsafe-to-chain-command
```

Specifically, it happens with the tests where I use `checkFieldValue` custom command:

```
    cy.checkFieldValue(fieldName, expectedValue).then(isMatched => {
      expect(isMatched).to.be.true;
    });
```

The rule can be helpful in certain scanerios, but when it comes to custom commands, it can feel limiting. I've addressed all potential scenarios within my custom command, so encountering a null or undefined value isn't an issue in this context. It simply signifies that the custom command encountered an error.

turadg: I'm reluctant to approve a lint suppression, especially in a tool I haven't used myself.

Could you accomplish this another way? Perhaps by making the expectation within the `checkFieldValue`? By the name it sounds like it would do the `expect` itself.

turadg: Oh, I see you do have it throwing now. So why ever return `true`? Just remove the return value and stop reading it here. Consider also renaming for clarity

rabi-siddique: Yeah, it makes it simpler. Updated

turadg (approved):

PR #285, opened 2024-05-22
ci: add scheduled runs of liquidation tests

rabi-siddique:
Due to the lengthy execution time of 20 to 25 minutes for liquidation tests on the local chain, the PR excludes these tests from PR executions. Instead, we will schedule regular runs of these tests.


frazarshad (approved):

toliaqat:
@rabi-siddique we should add a badge in readme to show the status of these tests and also set alerting if not already.

PR #286, opened 2024-05-22
tests: add tests for setting up governance params

rabi-siddique:
The PR does the following:
- Setup manager params from econ-gov.
- Setup auctioneer params from econ-gov. 
- Changes in setting up wallet test case for dapp-inter. 
- Exporting `AGORIC_NET` when running shell commands. This will assist when setting up these tests with emerynet.

### DIFF
--- a/test/e2e/specs/liquidation-reconstitution.spec.js
+++ b/test/e2e/specs/liquidation-reconstitution.spec.js
@@ -5,6 +5,7 @@ import {
   LIQUIDATING_TIMEOUT,
   LIQUIDATED_TIMEOUT,
   econGovURL,
+  MINUTE_MS,
 } from '../test.utils';
 
 describe('Wallet App Test Cases', () => {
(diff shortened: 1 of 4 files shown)


test/e2e/specs/liquidation.spec.js
```diff
@@ -21,23 +31,226 @@ describe('Wallet App Test Cases', () => {
...
+      cy.get('button').contains('Vote').click();
+      cy.get('p').contains('YES').click();
+      cy.get('input:enabled[value="Submit Vote"]').click();
+
+      cy.confirmTransaction();
+      cy.get('p').contains('sent').should('be.visible');
+    });
+
+    it('should wait for proposal to pass', () => {
+      cy.wait(60000 - Date.now() + startTime);
```
frazarshad, line 134: we can use MINUTE_MS

test/e2e/specs/liquidation-reconstitution.spec.js
```diff
@@ -21,23 +30,226 @@ describe('Wallet App Test Cases', () => {
...
+          cy.get('input').type('1');
+        });
+
+      cy.get('label')
+        .contains('MintFee')
+        .parent()
+        .within(() => {
+          cy.get('input').clear();
+          cy.get('input').type('0.5');
+        });
```
frazarshad, line 89: can this be turned into a custom command? feels like we've used this 100 times in this file

frazarshad (approved):

PR #290, opened 2024-05-24
ci: trigger liquidation tests manually

rabi-siddique:


.github/workflows/reusable-workflow.yml
```diff
@@ -42,7 +45,7 @@ jobs:
         env:
           # conditionals based on github event
           SYNPRESS_PROFILE: ${{ inputs.is_emerynet_test && 'daily-tests' || 'synpress' }}
-          CYPRESS_AGORIC_NET: ${{ inputs.is_emerynet_test && 'emerynet' || 'local' }}
+          CYPRESS_AGORIC_NET: ${{ inputs.AGORIC_NET }}
```
frazarshad, line 48: this condition is based on more than the value of inputs.AGORIC_NET it is also dependant on the github.event_name for some workflows such as  the vaults.yml workflow

.github/workflows/vaults.yml
```diff
@@ -9,7 +9,7 @@ on:
     - cron: '0 0 * * *'
   workflow_dispatch:
     inputs:
-      network:
+      AGORIC_NET:
```
frazarshad, line 12: lets follow the previous convention of non-capitalized input names

.github/workflows/liquidation-reconstitution.yml
```diff
@@ -17,3 +27,4 @@ jobs:
         docker-compose -f test/e2e/docker-compose-reconstitution.yml \
         --profile synpress up --build \
         --exit-code-from synpress
+      AGORIC_NET: ${{ inputs.network }}
```
frazarshad, line 30: why not leave the other code as is? we can just set this condition to: 
`${{  inputs.network === 'emerynet' }}`

frazarshad (approved):

PR #292, opened 2024-05-27
add missing tests and improvements for liquidation scenarios

rabi-siddique:
The PR does the following:
- Ensures the liquidation test cases don't timeout with emerynet. 
- Chore adjusting liquidation test cases to work consistently with emerynet and local chain. 
- Adds missing test cases related to collecting remaining collateral(Liquidation Happy Path). 
- Add test cases for verifying vaults are reconstitution(The auction numbers).
- Adds test cases to ensure all the vaults created during tests are closed. 
- Ensures using system executables for `agops` instead of the ones present in `node_modules`.


test/e2e/specs/liquidation-reconstitution.spec.js
```diff
@@ -435,25 +448,69 @@ describe('Wallet App Test Cases', () => {
...
 
         cy.verifyAuctionData(propertyName, expectedValue);
       });
+
+      it('should close the 100 debt vault and approve the transaction successfully', () => {
+        cy.skipWhen(AGORIC_NET === networks.LOCAL);
+        const regexVault100 = new RegExp('100(\\.\\d+)?');
+        cy.contains(regexVault100, { timeout: LIQUIDATING_TIMEOUT }).click();
+        cy.contains('Close Out Vault').click();
+        cy.contains('button.bg-interPurple', 'Close Out Vault').click();
```
frazarshad, line 490: we should avoid using css classes in our selectors, since they can be fragile

rabi-siddique: This happens when we close a vault. And we have two `Close Out Vault` buttons on the screen.

frazarshad (approved):

PR #295, opened 2024-05-28
chore(deps): remove turbo

rabi-siddique:


frazarshad (commented):
run `yarn` to generate the yarn.lock file and commit that as well

rabi-siddique:
> run `yarn` to generate the yarn.lock file and commit that as well

Thanks for pointing it out. Done

frazarshad (approved):

PR #296, opened 2024-05-28
cleanup tests for emerynet 

rabi-siddique:
The PR does the following:
- Implements a test to reset the ATOM price to `12.34` in Emerynet, ensuring the resumption of normal operations.
- Adds a test to remove the `150 IST` bid in the liquidation happy path tests, to prevent any impact on subsequent test runs.

### DIFF
--- a/test/e2e/specs/liquidation-reconstitution.spec.js
+++ b/test/e2e/specs/liquidation-reconstitution.spec.js
@@ -511,6 +511,11 @@ describe('Wallet App Test Cases', () => {
           );
         });
       });
+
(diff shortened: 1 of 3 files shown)


frazarshad:
what if the tests fail, due to flakiness and the values arent reset?

rabi-siddique:
> what if the tests fail, due to flakiness and the values arent reset?

I have tested it locally with emerynet. There were no issues of flakiness or values not being reset. Assuming that the values are not reset in some cases, it won't give a problem ==> I think it will result in some test cases relevant to auction failure and that's it.

frazarshad (approved):

PR #297, opened 2024-06-03
ci: artifacts setup for dapp-inter

rabi-siddique:
This pull request enables the uploading of end-to-end tests to Google Cloud Platform (GCP).

I tested this workflow in this [GitHub Action run](https://github.com/Agoric/dapp-inter/actions/runs/9352479080/job/25740739716?pr=297) by making a dummy commit.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,4 @@
 node_modules/
 docker
+# Ignore generated credentials from google-github-actions/auth
+gha-creds-*.json
--- a/.github/workflows/reusable-workflow.yml
+++ b/.github/workflows/reusable-workflow.yml
@@ -35,9 +35,20 @@ jobs:
   e2e:
     runs-on: ubuntu-latest
 
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
+
+      - name: 'GCP auth'
+        uses: 'google-github-actions/auth@v2'
+        if: ${{ inputs.is_emerynet_test }}
+        with:
+          project_id: 'simulationlab'
+          workload_identity_provider: 'projects/60745596728/locations/global/workloadIdentityPools/github/providers/dapp-inter'
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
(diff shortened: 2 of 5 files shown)


.dockerignore
```diff
@@ -1,2 +1,4 @@
 node_modules/
 docker
+# Ignore generated credentials from google-github-actions/auth
+gha-creds-*.json
```
LuqiPan, line 4: _question_

Why did we need to git ignore this file given it's only generated when running inside a github action?

.github/workflows/reusable-workflow.yml
```diff
@@ -89,13 +100,25 @@ jobs:
           CYPRESS_BIDDER_MNEMONIC_INPUT: ${{ inputs.bidder_mnemonic }}
           CYPRESS_BIDDER_ADDRESS_INPUT: ${{ inputs.bidder_address }}
 
-      - name: Archive e2e artifacts
+      - name: Archive e2e artifacts locally
         uses: actions/upload-artifact@v3
```
LuqiPan, line 104: _low priority_

Bump this action to v4 as v3 will be deprecated in half a year?

.github/workflows/reusable-workflow.yml
```diff
@@ -89,13 +100,25 @@ jobs:
           CYPRESS_BIDDER_MNEMONIC_INPUT: ${{ inputs.bidder_mnemonic }}
           CYPRESS_BIDDER_ADDRESS_INPUT: ${{ inputs.bidder_address }}
 
-      - name: Archive e2e artifacts
+      - name: Archive e2e artifacts locally
```
LuqiPan, line 103: _low priority_

I don't think `locally` is accurate. I'd call it `Upload e2e artifacts as workflow artifacts` or something similar

rabi-siddique: Makes sense. Updated

.github/workflows/reusable-workflow.yml
rabi-siddique: You're right. Done

.dockerignore
rabi-siddique: You're right. It was mentioned in the prereqs [over here](https://github.com/marketplace/actions/authenticate-to-google-cloud#prerequisites)

.github/workflows/reusable-workflow.yml
```diff
@@ -89,13 +100,25 @@ jobs:
           CYPRESS_BIDDER_MNEMONIC_INPUT: ${{ inputs.bidder_mnemonic }}
           CYPRESS_BIDDER_ADDRESS_INPUT: ${{ inputs.bidder_address }}
 
-      - name: Archive e2e artifacts
-        uses: actions/upload-artifact@v3
-        if: always()
+      - name: Upload e2e artifacts as workflow artifacts
+        uses: actions/upload-artifact@v4
+        if: ${{ !inputs.is_emerynet_test }}
```
frazarshad, line 105: we should be adding `!cancelled()` here as well. this will allow this step to run even if the previous step (the running of test) fails.
```suggestion
        if: ${{ !inputs.is_emerynet_test && !cancelled() }}
```

frazarshad (approved):

.dockerignore
rabi-siddique: Sure

rabi-siddique: Done. [LINK](https://github.com/Agoric/dapp-econ-gov/pull/105)

PR #298, opened 2024-06-03
chore: update adding keys command

rabi-siddique:


test/e2e/support.js
```diff
@@ -9,7 +9,9 @@ Cypress.Commands.add('addKeys', params => {
   const { keyName, mnemonic, expectedAddress } = params;
   const command = `echo ${mnemonic} | agd keys add ${keyName} --recover ${keyRing}`;
 
-  cy.exec(command).then(({ stdout }) => {
+  cy.exec(command, {
+    failOnNonZeroExit: false,
+  }).then(({ stdout }) => {
```
frazarshad, line 14: can you update the PR description on why this change is needed?

frazarshad: also wont this flag also fail to throw an error if the command fails? resulting in the test falsely passing?

rabi-siddique: No. The test cases will fail if the command fails. The change is implemented to hide stack-traces.

frazarshad (approved):

Issue #299, opened 2024-06-04
Chrome incorrectly flagging underlying IPFS hash as 'dangerous site' when you attempt to access via app.inter.trade


otoole-brendan:
## Describe the bug

Chrome is flagging the .dweb.link-appended IPFS hash as 'Dangerous' as of Monday 3rd June.

## To Reproduce

1. Go to app.inter.trade
2. See resolution to IPFS hash
3. See error/dangerous site message

## Expected behavior

Should not be flagged as dangerous

## Screenshots

![image](https://github.com/Agoric/dapp-inter/assets/93355546/8be7e20d-4ac4-406f-a3d4-521c833297eb)

otoole-brendan:
Quickest way to resolve is for EC to push new hash which requires a dapp-inter change before DCF can deploy to IPFS and get a new hash. @samsiegart assigning to you to complete that action.

PR #300, opened 2024-06-04
feat: trivially change price feed error message

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/299

### DIFF
--- a/src/service/vaults.ts
+++ b/src/service/vaults.ts
@@ -74,7 +74,7 @@ const watchPriceFeeds = (prefix: string) => {
               .getState()
               .setPriceError(
                 brand,
-                new Error('Unexpected error, price feed missing'),
+                new Error('Unexpected error, price feed missing.'),
               );
           }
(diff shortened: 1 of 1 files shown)


otoole-brendan (approved):

src/service/vaults.ts
```diff
@@ -74,7 +74,7 @@ const watchPriceFeeds = (prefix: string) => {
               .getState()
               .setPriceError(
                 brand,
-                new Error('Unexpected error, price feed missing'),
+                new Error('Unexpected error, price feed missing.'),
```
erights, line 77: For want of a dot, the hash was lost

samsiegart: To be clear for posterity, a new hash was the [goal](https://github.com/Agoric/dapp-inter/issues/299#issuecomment-2147990966) (I couldn't find anywhere to put a nail 😄)

PR #301, opened 2024-06-07
passing env variables to the reusable workflow

rabi-siddique:
While running emerynet tests in CI for liquidation workflows, the environment variables were not picked up in the code. Their values were `undefined`.

After inspecting, I found that we need to pass the secrets to the reusable workflow from where it's being called. You can refer to this [GitHub documentation on using inputs and secrets in a reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) for more details.

In our CI setup, we have a file named `reusable.workflow.yml`, which is called in both `liquidation.yml` and `liquidation-reconstitution.yml`. In this PR, I am passing the secrets from these workflows to the reusable workflow.

### DIFF
--- a/.github/workflows/liquidation-reconstitution.yml
+++ b/.github/workflows/liquidation-reconstitution.yml
@@ -48,3 +48,5 @@ jobs:
       user1_address: ${{ inputs.user1_address }}
       bidder_mnemonic: ${{ inputs.bidder_mnemonic }}
       bidder_address: ${{ inputs.bidder_address }}
+
+    secrets: inherit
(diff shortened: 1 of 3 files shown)


turadg (approved):
The change is an improvement.

I wonder how this could have been detected earlier. Maybe with some config validation in the jobs.

rabi-siddique:
> I wonder how this could have been detected earlier. Maybe with some config validation in the jobs.

That's a great point. Implementing configuration validation would be super helpful. I'll work on setting up the necessary configurations soon.

PR #307, opened 2024-06-11
chore: use different timeouts based on network

rabi-siddique:
The PR does the following:
- Makes sure we use different timeouts for liquidation based on network. 
- Modifications in tests to ensure we don't wait much for desired results in case of failure.

### DIFF
--- a/test/e2e/specs/liquidation.spec.js
+++ b/test/e2e/specs/liquidation.spec.js
@@ -1,7 +1,5 @@
 import {
   mnemonics,
-  LIQUIDATING_TIMEOUT,
-  LIQUIDATED_TIMEOUT,
   MINUTE_MS,
   networks,
   configMap,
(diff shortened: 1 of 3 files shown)


test/e2e/specs/liquidation.spec.js
```diff
@@ -477,7 +477,7 @@ describe('Wallet App Test Cases', () => {
       cy.skipWhen(AGORIC_NET === networks.LOCAL);
 
       const propertyName = 'book0.collateralAvailable';
-      const expectedValue = '9.659302 ATOM';
+      const expectedValue = '9.659301 ATOM';
```
frazarshad, line 480: can you explain why you had to update this value?

rabi-siddique: When matching auction values, the expected value for `collateralAvailable` is `9.659301 ATOM`. It is rarely `9.659302 ATOM`

frazarshad: if it can have two close possible values, i think a better approach would be to check that the difference is within a certain limit:
```
expect(Math.abs(expectedValue - foundValue)).toBeLessThan(0.0001);
```

rabi-siddique: It is a generic function that compares string values. I don't want to mess it up because it is being used in multiple places. 
`expect(Math.abs(expectedValue - foundValue)).toBeLessThan(0.0001);` incorrect as I said we compare strings. Strings have token denominations as well.

mudassir-agoric (approved):

PR #309, opened 2024-06-11
test: updated tests to reduce flakiness for wallet setup

frazarshad:
This PR makes the following updates to CI tests to reduce flakiness:

- first it updates the flow to get the chain address. we were originally using the wallet dapp for this but a good solution instead is to use the address of the original 'Agoric' chain, since it will be the same for emerynet
- second, it updates the flow for provisoin tokens using the emerynet faucet. previously we were doing it through the emerynet faucet ui, but now we are performing it thorugh direct API request, which allows us to retry on failure

### DIFF
--- a/test/e2e/specs/test.spec.js
+++ b/test/e2e/specs/test.spec.js
@@ -24,44 +24,9 @@ describe('Vaults UI Test Cases', () => {
 
         cy.getWalletAddress('Agoric').then(address => {
           // provision BLD
-          cy.request({
-            method: 'POST',
-            url: 'https://emerynet.faucet.agoric.net/go',
-            body: {
-              address,
-              command: 'delegate',
-              clientType: 'SMART_WALLET',
-            },
-            headers: {
-              Accept:
-                'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7',
-              'Content-Type': 'application/x-www-form-urlencoded',
-            },
-            timeout: 4 * MINUTE_MS,
-            retryOnStatusCodeFailure: true,
-          }).then(resp => {
-            expect(resp.body).to.eq('success');
-          });
-
+          cy.provisionFromFaucet(address, 'delegate');
           // provision IST
-          cy.request({
-            method: 'POST',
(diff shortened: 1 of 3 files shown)


test/e2e/specs/test.spec.js
```diff
@@ -16,46 +16,53 @@ describe('Vaults UI Test Cases', () => {
...
         });
       } else {
-        const walletAddress = {
-          value: null,
-        };
         cy.setupWallet({
           createNewWallet: true,
           walletName: 'my created wallet',
+          selectedChains: ['Agoric'],
         });
```
rabi-siddique, line 23: I am assuming this is done to get a wallet address?

frazarshad: yes

test/e2e/specs/test.spec.js
```diff
@@ -16,46 +16,53 @@ describe('Vaults UI Test Cases', () => {
...
-        cy.acceptAccess();
-
-        cy.origin('https://wallet.agoric.app/', () => {
-          cy.visit('/wallet/');
+        cy.getWalletAddress('Agoric').then(address => {
+          // provision BLD
+          cy.request({
+            method: 'POST',
+            url: 'https://emerynet.faucet.agoric.net/go',
+            body: {
```
rabi-siddique, line 30: Perhaps add the url as a const in `test.utils.js`

test/e2e/specs/test.spec.js
```diff
@@ -16,46 +16,53 @@ describe('Vaults UI Test Cases', () => {
...
+            },
+            headers: {
+              Accept:
+                'text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7',
+              'Content-Type': 'application/x-www-form-urlencoded',
+            },
+            timeout: 4 * MINUTE_MS,
+            retryOnStatusCodeFailure: true,
+          }).then(resp => {
+            expect(resp.body).to.eq('success');
```
rabi-siddique, line 63: Given the duplication in the code, I think you should create a function or a custom command taking `command` as an argument.

rabi-siddique (approved):

PR #310, opened 2024-06-11
chores in cleanup tests for liquidation 

rabi-siddique:
The PR does the following:
- Adds waits when verifying auction values to ensure accuracy.
- Segregates tests related to claiming collateral into three separate tests for each vault. This approach has proven to be more reliable, as looping introduces an element of flakiness due to the rapid processing.

### DIFF
--- a/test/e2e/specs/liquidation.spec.js
+++ b/test/e2e/specs/liquidation.spec.js
@@ -428,28 +428,28 @@ describe('Wallet App Test Cases', () => {
     it('should verify the value of startPrice from the CLI successfully', () => {
       const propertyName = 'book0.startPrice';
       const expectedValue = '9.99 IST/ATOM';
-      cy.wait(3000);
+      cy.wait(3000); // eslint-disable-line cypress/no-unnecessary-waiting
       cy.verifyAuctionData(propertyName, expectedValue);
     });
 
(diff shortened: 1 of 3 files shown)


test/e2e/specs/liquidation.spec.js
```diff
@@ -428,28 +428,28 @@ describe('Wallet App Test Cases', () => {
     it('should verify the value of startPrice from the CLI successfully', () => {
       const propertyName = 'book0.startPrice';
       const expectedValue = '9.99 IST/ATOM';
-
+      cy.wait(3000); // eslint-disable-line cypress/no-unnecessary-waiting
       cy.verifyAuctionData(propertyName, expectedValue);
```
frazarshad, line 432: cant we just add this line inside cy.verifyAuctionData as well?

rabi-siddique: We can do that. But it makes `verifyAuctionData` fail for some reason. I did not dig much into but perhaps I will then to `cy.wait().then()` in verifyAuctionData. Perhaps

rabi-siddique: Yes, it is required there

frazarshad: please update it as well

frazarshad (approved):

Issue #311, opened 2024-06-11
Total # of active vaults not including vaults with zero IST debt


otoole-brendan:
## Describe the bug

App.inter.trade shows `Total # of active vaults `which uses sum of `numActiveVaults`. However it does not include vaults with zero IST debt and therefore is technically incorrect and underreporting the number of active vaults.

Info.inter.trade accounts for these vaults and includes them in the count. 

Update dapp-inter to use this query to return the correct value ([Slack reference](https://agoricopco.slack.com/archives/C038TM0EUKG/p1718062228425269?thread_ts=1718057067.747099&cid=C038TM0EUKG))

curl -s 'https://api.subquery.network/sq/agoric-labs/agoric-mainnet-v2' -H 'Content-Type: application/json' --data-binary '{
  "query": "query {\tvaults(filter: {state: {equalTo: \"active\"}}) {totalCount}}"
}' | jq .

## To Reproduce

1. Compare $ of active vaults visible on app.inter.trade with that of info.inter.trade

## Expected behavior

Should in include vaults that have zero IST debt

samsiegart:
@otoole-brendan Do we have subquery indexers for testnets as well? At the very least, we won't have it for local networks. I can make the dapp use this subquery data for mainnet, and all other networks it would still use the same old data if that works.

otoole-brendan:
> [@otoole-brendan](https://github.com/otoole-brendan) Do we have subquery indexers for testnets as well? At the very least, we won't have it for local networks. I can make the dapp use this subquery data for mainnet, and all other networks it would still use the same old data if that works.

that's fine @samsiegart ! @toliaqat can answer about indexers for testnets - I'm not sure

PR #315, opened 2024-06-13
fix: use subquery to count all active vaults

samsiegart:
fixes https://github.com/Agoric/dapp-inter/issues/311

### DIFF
--- a/src/components/MainContentWrapper.tsx
+++ b/src/components/MainContentWrapper.tsx
@@ -6,8 +6,9 @@ import {
 } from '@agoric/zoe/src/contractSupport';
 import { useAtomValue } from 'jotai';
 import { PropsWithChildren, useMemo } from 'react';
-import { displayFunctionsAtom } from 'store/app';
+import { displayFunctionsAtom, networkConfigAtom } from 'store/app';
 import { useVaultStore } from 'store/vaults';
+import { useTotalActiveVaultsQuery } from 'utils/hooks';
 
 type TickerItemProps = {
   label: string;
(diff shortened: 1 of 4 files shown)


src/components/MainContentWrapper.tsx
```diff
@@ -72,11 +75,14 @@ const MainContentWrapper = ({ children }: PropsWithChildren) => {
...
         let totalDebt = AmountMath.makeEmpty(debtBrand);
         let totalLocked = AmountMath.makeEmpty(debtBrand);
-        let numVaults = 0;
+        let numVaultsWithDebt = 0;
 
         [...metrics.values()].forEach(
           ({ numActiveVaults, totalCollateral, totalDebt: managerDebt }) => {
-            numVaults += numActiveVaults;
+            // This metric misleadingly only counts vaults with non-zero debt,
+            // even if they're 'active'.
```
turadg, line 83: UNTIL https://github.com/Agoric/agoric-sdk/issues/9463

(closed rn but it will probably happen eventually)

src/utils/hooks.ts
```diff
@@ -54,3 +55,36 @@ export const useAuctionTimer = (schedule: LiquidationSchedule | null) => {
...
+
+/**
+ * Polls the total number of active vaults if configured to use mainnet,
+ * otherwise returns null.
+ *
+ * This fetches from subquery, rather than a vstorage query to a network node,
+ * because the query is more complex. However, subquery only works on mainnet,
+ * hence why it checks the network config.
+ */
+export const useTotalActiveVaultsQuery = (networkConfigUrl: string) => {
```
turadg, line 67: please mention that it's expected to lag no more than 15 minutes.

should that go in the UI somewhere?

turadg (approved):

src/components/MainContentWrapper.tsx
samsiegart: Added the issue link

src/utils/hooks.ts
samsiegart: Might be a good idea, I'll ask @otoole-brendan but won't block this PR on it

otoole-brendan: don't think the lag disclaimer is necessary @turadg - this metric is informational only to give users some high-level stats. That it's not immediately updated is an acceptable experience for what it's designed for

PR #333, opened 2024-07-26
Commit DEPLOYED_HASH to deploy branch instead of main

iomekam:
We can have the `DEPLOYED_HASH` value stored in the deploy branch. We already make a commit on top of the target commit, so it's a simple addition.

Passing actions run: https://github.com/Agoric/dapp-inter/actions/runs/10115904779

### DIFF
--- a/.github/workflows/deploy-poller.yml
+++ b/.github/workflows/deploy-poller.yml
@@ -54,25 +54,41 @@ jobs:
         if: env.SKIP_RUN == 'false'
         run: |
           # Define the URL from the environment variable
-          URL="${{ vars.CHAIN_URL }}/agoric/vstorage/data/published.vaultFactory.governance"
+          URL="${{ env.CHAIN_URL }}/agoric/vstorage/data/published.vaultFactory.governance"
 
           # Fetch the new value
           NEW_REFERENCED_UI=$(curl -s $URL | jq -r '.value' | jq -r '.values[]' | jq -r .body | sed 's/^#//' | jq -r .current.ReferencedUI.value)
           if [ ${{ github.event.inputs.COMMIT_URL }} ]; then
             NEW_REFERENCED_UI="${{ github.event.inputs.COMMIT_URL }}"
           fi
+          
+          # Check if the deploy branch exists on the remote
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-poller.yml
```diff
@@ -54,26 +54,42 @@ jobs:
         if: env.SKIP_RUN == 'false'
         run: |
           # Define the URL from the environment variable
-          URL="${{ vars.CHAIN_URL }}/agoric/vstorage/data/published.vaultFactory.governance"
+          URL="${{ env.CHAIN_URL }}/agoric/vstorage/data/published.vaultFactory.governance"
```
samsiegart, line 57: What's the difference between env and vars?

.github/workflows/deploy-poller.yml
```diff
@@ -23,8 +23,8 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
-      CHAIN_URL: ${{ vars.CHAIN_URL || github.event.inputs.CHAIN_URL }}
-      DEPLOY_BRANCH: ${{ vars.DEPLOY_BRANCH || github.event.inputs.DEPLOY_BRANCH }}
+      CHAIN_URL: ${{ github.event.inputs.CHAIN_URL || vars.CHAIN_URL }}
```
samsiegart, line 26: This looks like a fallback. In which scenario would we expect it to get the value from `github.event.inputs` as opposed to `vars`?

iomekam: When trying to run the workflow manually
Screenshot 2024-07-26 at 1 55 08 PM

.github/workflows/deploy-poller.yml
iomekam: There are three things here:

Repository environment variables
Environment variables within the workflow
Inputs within the workflow.

`vars.*` refers to repository environment variables. These are values that should be consistent across any workflow run in your repo

`github.event.inputs.*` refers to inputs passed into the workflow

`env.*` refers to environment variables created and available within your workflow

samsiegart (approved):

PR #353, opened 2024-08-26
ci: input tags for a3p

rabi-siddique:
The PR does the following:
- Adding support to test the a3p chain with custom tags.
- Removes the input of governance wallets in ci.

### DIFF
--- a/.github/workflows/liquidation-reconstitution.yml
+++ b/.github/workflows/liquidation-reconstitution.yml
@@ -32,22 +32,6 @@ on:
         description: 'bidder account address'
         required: false
         type: string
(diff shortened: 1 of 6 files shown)


frazarshad (approved):

Muneeb147 (approved):
Looks good. Were `gov1/2` dispatch inputs un-used hence removed?

rabi-siddique:
> Looks good. Were `gov1/2` dispatch inputs un-used hence removed?

Removing them primarily because GitHub might reveal them when debug logging is enabled. Considering the significance of these wallets, we've chosen to keep them in `secrets` instead of `inputs`.

PR #363, opened 2024-08-29
Faucet Polling in E2E Tests

usmanmani1122:
## Description
Updated test to poll for transaction completion

### DIFF
--- a/test/e2e/support.js
+++ b/test/e2e/support.js
@@ -3,9 +3,7 @@ import {
   networks,
   configMap,
   FACUET_HEADERS,
-  MINUTE_MS,
   agoricNetworks,
-  FAUCET_URL_MAP,
 } from './test.utils';
 
 const AGORIC_NET = Cypress.env('AGORIC_NET') || 'local';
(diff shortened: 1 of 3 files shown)


test/e2e/support.js
```diff
@@ -176,20 +174,41 @@ Cypress.Commands.add('connectWithWallet', (options = {}) => {
...
+      .request({
+        method: 'GET',
+        url: `https://${AGORIC_NET}.faucet.agoric.net/api/transaction-status/${txHash}`,
+      })
+      .then(resp => {
+        const { transactionStatus } = resp.body;
+        if (transactionStatus === TRANSACTION_STATUS.NOT_FOUND)
+          // eslint-disable-next-line cypress/no-unnecessary-waiting
+          return cy.wait(2000).then(() => getStatus(txHash));
+        else return cy.wrap(transactionStatus);
```
rabi-siddique, line 194: Should we use a maximum number of retires or some timeout to avoid an infinite loop? Is it possible that `return cy.wait(2000).then(() => getStatus(txHash));` continues indefinitely?

usmanmani1122: Cypress commands have a default [timeout](https://docs.cypress.io/guides/references/configuration#Timeouts) of 4 seconds so unless we have changed it, this shouldn't happen

rabi-siddique: Command timeout is 6 minutes for testnets and 2 minutes with local chain. Would that be fine? I think we should have 4 or 5 retries. That would be enough to make the call invalid, I guess.

usmanmani1122: I don't think we can confidentaly say that `N` number of retries should be enough to include a transaction in a block. I would say let's keep trying until the Cypress command timeout exceeds instead to get maximum retires in. What do you think?

rabi-siddique: Okies. Let's see how tests perform with this change.

rabi-siddique (approved):

test/e2e/support.js
usmanmani1122: Yes you have a  valid point. But I would like to see how often do we timeout waiting for the transaction to complete. Maybe we will need to do something about that if it is frequent.

PR #367, opened 2024-09-17
test: pause old auctioneer to avoid clobbering vstorage

dckc:
DRAFT until:

 - [ ] figure out how to integrate [auctionHistory.js](https://gist.github.com/dckc/9180ee35518cb9bbe14bd63c24efdfd8#file-auctionhistory-js) core eval script

I also haven't managed to run the liquidation tests all the way thru.

### DIFF
--- a/test/e2e/specs/liquidation.spec.js
+++ b/test/e2e/specs/liquidation.spec.js
@@ -195,6 +195,12 @@ describe('Wallet App Test Cases', () => {
     });
   });
 
+  context('Old auctioneer', () => {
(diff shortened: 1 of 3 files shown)


rabi-siddique:
> I also haven't managed to run the liquidation tests all the way thru.

You can include the label `liquidation-testing` in this PR and it will trigger the liquidation e2e tests. 
The same applies to liquidation reconstitution e2e tests when the label `liquidation-reconstitution` is added to a PR.

PR #380, opened 2024-10-11
chore: compare values upto 2 decimal places

rabi-siddique:
This PR ensures that the expected value for shortfall balance comparisons is rounded to two decimal places, matching the format of the actual test output.

Currently, the comparison fails because the actual value `5.53` does not match the expected value `5.525`.


amessbee (approved):
looks like you know what you are doing!

PR #386, opened 2024-10-22
test: fixed flake due to smart wallet required dialog

frazarshad:


test/e2e/support.js
```diff
@@ -194,6 +194,7 @@ const connectWalletLocalChain = ({ isVaultsTests = false }) => {
 };
 
 const connectWalletTestnet = () => {
+  cy.contains('button', 'Dismiss').click();
   cy.get('button').contains('Local Network').click();
```
rabi-siddique, line 198: I think this might trouble with Liquidation tests.

rabi-siddique: It's my bad for not realizing we're using the same function in Vault tests.

rabi-siddique (approved):

rabi-siddique:
LGTM. The one test case failing in liquidation tests is not related to the changes in this PR.

Issue #388, opened 2024-12-04
Add dATOM icon to Dapp-inter and dATOM default for leap elements


otoole-brendan:
## What is the Problem Being Solved?

As part of the collateral onboarding process for dATOM - we need to add the dATOM icon to dapp UI so it renders throughout.

## Description of the Design

1. Configure dapp-inter UI to render dATOM icon from Cosmos chain registry https://github.com/cosmos/chain-registry/blob/master/neutron/images/dATOM.svg 

1. Add dATOM default to Leap Elements

## Test Plan
Visually confirm in UI

PR #389, opened 2024-12-05
feat: add dATOM support

samsiegart:
refs https://github.com/Agoric/dapp-inter/issues/388

## Screenshots

Screen Shot 2024-12-04 at 10 57 07 PM
Screen Shot 2024-12-04 at 11 07 05 PM

## Problem

The leap elements modal says "failed to get IBC support for neutron to agoric".  See screenshot:

Screen Shot 2024-12-06 at 9 02 16 AM

My first thought was to maybe update leap elements, but leap dashboard has the same issue:

Screen Shot 2024-12-04 at 11 02 00 PM

I thought maybe we could link to https://go.skip.build/ as a fallback, but that doesn't seem to support neutron to agoric either.

I'm guessing we need to add the IBC info to https://github.com/cosmos/chain-registry/tree/master/_IBC, but it's unclear if that will directly fix the issue without confirming with the leap team.


samsiegart:
It's not even possible to send dATOM to agoric with keplr yet either:
Screen Shot 2024-12-06 at 8 58 34 AM

0xpatrickdev (approved):

PR #401, opened 2025-05-28
feat!: ist winddown

samsiegart:
Just updates @agoric/web-components and fixes type errors so it builds. This will cause it to use BLD as the fee token in the wallet connection. We don't need to mess with the smart wallet provision flow because the mint limits are 0 so that code path shouldn't happen.

Related: https://github.com/Agoric/dapp-psm/pull/175


src/store/adjustVault.ts
```diff
@@ -7,7 +7,7 @@ import {
 import {
   floorMultiplyBy,
   ratioGTE,
-} from '@agoric/zoe/src/contractSupport/ratio';
+} from '@agoric/zoe/src/contractSupport/index';
```
turadg, line 10: this is in ERTP now. you might not need `@agoric/zoe` anymore

```suggestion
import { floorMultiplyBy, ratioGTE } from '@agoric/ertp/src/ratio.js';
```

src/store/createVault.ts
```diff
@@ -8,9 +8,9 @@ import {
   istAvailable,
 } from 'utils/vaultMath';
 import { pursesAtom } from './app';
-import { ratioGTE } from '@agoric/zoe/src/contractSupport/ratio';
-import { AmountMath } from '@agoric/ertp';
-import type { Amount, NatValue } from '@agoric/ertp/src/types';
+import { ratioGTE } from '@agoric/zoe/src/contractSupport/index';
+import { AmountMath } from '@agoric/ertp/src/index';
```
turadg, line 12: from `@agoric/ertp` didn't work?

yarn.lock
```diff
@@ -53,6 +53,19 @@
     jsesc "^2.5.1"
     source-map "^0.5.0"
 
+"@agoric/base-zone@^0.1.1-u20.0":
```
turadg, line 56: this file got much larger but shouldn't. please adjust the versions. you could use `*` for '@endo/' versions so that the agoric deps imply them. the agoric deps should also be linked up to a tag like u20.

turadg (changes requested):

src/store/adjustVault.ts
samsiegart: Reverted to older version of ertp so this won't be included

src/store/createVault.ts
samsiegart: It did, thanks, fixed

package.json
```diff
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "export NODE_OPTIONS='--max-old-space-size=8192' && vite build",
     "preview": "vite preview",
+    "lint:types": "echo 'XXX tsc is not working, even though vite build is successful' && tsc --noEmit",
```
turadg, line 10: nice solution

src/components/VaultSummary.tsx
```diff
@@ -10,7 +10,8 @@ import { calculateCurrentDebt } from '@agoric/inter-protocol/src/interest-math';
 import {
   ceilMultiplyBy,
   makeRatioFromAmounts,
-} from '@agoric/zoe/src/contractSupport';
+  multiplyBy,
+} from '@agoric/zoe/src/contractSupport/ratio';
```
turadg, line 14: these are all in `@agoric/ertp` now. since this app is going away I won't push for it

turadg (approved):

